PR #1, opened 2020-11-11
MVP for alertmanager charm

dstathis:
This seems to be working perfectly. More tests coming soon.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,7 @@
 [flake8]
 max-line-length = 99
 select: E,W,F,C,N
+ignore = W503
 exclude:
   venv
   .git
--- a/run_tests
+++ b/run_tests
@@ -12,8 +12,8 @@ else
     export PYTHONPATH="src:$PYTHONPATH"
 fi
 
-flake8
-black --diff
+flake8 || exit $?
+black --diff --check --skip-string-normalization src tests || exit $?
 which pytest > /dev/null
 if [ $? -eq 0 ]; then
     python3 -m pytest -v "$@"
--- a/run_tests
+++ b/run_tests
@@ -13,6 +13,7 @@ else
 fi
 
 flake8
+black --diff
 which pytest > /dev/null
 if [ $? -eq 0 ]; then
     python3 -m pytest -v "$@"
--- a/src/charm.py
+++ b/src/charm.py
@@ -31,8 +31,8 @@ receivers:
      service_key: '{pagerduty_key}'
 """
 
-HA_PORT = '9094'
-UNIT_ADDRESS = '{}-{}.{}-endpoints.{}.svc.cluster.local'
+HA_PORT = "9094"
+UNIT_ADDRESS = "{}-{}.{}-endpoints.{}.svc.cluster.local"
 
 
 class BlockedStatusError(Exception):
@@ -46,18 +46,24 @@ def status_catcher(func):
             func(self, *args, **kwargs)
         except BlockedStatusError as e:
             self.unit.status = BlockedStatus(str(e))
+
     return new_func
 
 
 class AlertmanagerCharm(CharmBase):
-
     def __init__(self, *args):
-        log.debug('Initializing charm.')
+        log.debug("Initializing charm.")
         super().__init__(*args)
         self.framework.observe(self.on.config_changed, self.on_config_changed)
-        self.framework.observe(self.on['alerting'].relation_changed, self.on_alerting_changed)
-        self.framework.observe(self.on['alertmanager'].relation_changed, self.on_alertmanager_changed)
-        self.framework.observe(self.on['alertmanager'].relation_departed, self.on_alertmanager_departed)
+        self.framework.observe(
+            self.on["alerting"].relation_changed, self.on_alerting_changed
+        )
+        self.framework.observe(
+            self.on["alertmanager"].relation_changed, self.on_alertmanager_changed
+        )
+        self.framework.observe(
+            self.on["alertmanager"].relation_departed, self.on_alertmanager_departed
+        )
 
     @status_catcher
     def on_alerting_changed(self, event):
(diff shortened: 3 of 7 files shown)


src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
+        self.framework.observe(
+            self.on["alertmanager"].relation_changed, self.on_alertmanager_changed
+        )
+        self.framework.observe(
+            self.on["alertmanager"].relation_departed, self.on_alertmanager_departed
+        )
 
     @status_catcher
     def on_alerting_changed(self, event):
+        self.update_alerting(event.relation)
```
balbirthomas, line 70: Doesn't `on_alerting_changed` seem redundant if all you are doing is making a method call to another method ? Or would it be preferable to separate this workflow into two methods with distinct responsibilites
1. One that checks if using is leader and requests the other method to generate the list of alerting addresses. And then sets the so obtained relation data.
2. Another that only generates the alerting relation data.

Wouldn't it be preferable to put the `is_leader` check in this event relation handler, thereby making it immediately obvious that the event is handled only by the leader ? (and a similar comment for other such cases in this pull request).

src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
+    @status_catcher
+    def on_alertmanager_changed(self, event):
+        if self.unit.is_leader():
+            self.configure()
+            for relation in self.model.relations["alerting"]:
+                self.update_alerting(relation)
+
+    @status_catcher
+    def on_alertmanager_departed(self, event):
+        if self.unit.is_leader():
```
balbirthomas, line 101: Nit Pick: Inverting this would lead to more flat (as opposed to nested) code flow. So for instance if the check was `if not self.unit.is_leader()` then return. I do see you seem to have a preference for the other way round here and elsewhere in your pull request.

balbirthomas:
It is already possible to merge this pull request. I believe @jlounder recommendation is that for pull requests against Canonical master branch, github settings should be such that two approvals are mandatory.

src/charm.py
```diff
@@ -103,67 +145,74 @@ def build_pod_spec(self):
...
+            "containers": [
+                {
+                    "name": self.app.name,  # self.app.name is defined in metadata.yaml
+                    "imageDetails": image_details,
+                    "args": [
+                        "--config.file=/etc/alertmanager/alertmanager.yaml",
+                        "--storage.path=/alertmanager",
+                        f'--web.listen-address=:{config["port"]}',
+                    ]
+                    + ha_targets,
```
balbirthomas, line 170: This implies every time a new alertmanager unit is added, every alertmanager pod is restarted. Is this correct ? Does this mean alertmanager does not provide an API that allows dynamic configuation without pod restart ?

src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
-
         self.unit.status = ActiveStatus()
-        log.debug('Pod spec set successfully.')
+        log.debug("Pod spec set successfully.")
 
     def build_config_file(self):
         """Create the alertmanager config file from self.model.config"""
         if not self.model.config["pagerduty_key"]:
-            raise BlockedStatusError('Missing pagerduty_key config value')
+            raise BlockedStatusError("Missing pagerduty_key config value")
```
balbirthomas, line 136: I am not sure what the requirements are. Is it the case that we do not allow use of alertmanager without pager duty ? Or is it the case we plan to support alternative alerting mechanisms, one of which is pagerduty and that at least one alerting mechanism must be selected ?

.flake8
```diff
@@ -1,6 +1,7 @@
 [flake8]
 max-line-length = 99
 select: E,W,F,C,N
+ignore = W503
```
chipaca, line 4: the flip-flopping on W503 is/was annoying, yes

src/charm.py
chipaca: FWIW the nested approach is more python-ish and the non-nested one is more Go-ish :-)

src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
+            for i in range(num_units):
+                addrs.append(
+                    UNIT_ADDRESS.format(
+                        self.meta.name, i, self.meta.name, self.model.name
+                    )
+                )
+            if addrs != json.loads(relation.data[self.app].get("addrs", "null")):
+                relation.data[self.app]["addrs"] = json.dumps(addrs)
+
+    @status_catcher
```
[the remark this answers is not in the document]
chipaca: I like this, and I'd like operator itself to have status exceptions that DTRT (without needing a decorator). I see the decorator as an implementation detail, a necessary evil even to achieve the wanted 'just raise an exception to set a failing status'

chipaca (approved):
LGTM for a first approach. Lots of things to improve (some of which I've noted below), none of them blockers (ie fine to do in followups).

src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
+    def on_alertmanager_departed(self, event):
+        if self.unit.is_leader():
+            self.configure()
+            for relation in self.model.relations["alerting"]:
+                self.update_alerting(relation)
 
     @status_catcher
     def on_config_changed(self, _):
+        self.configure()
```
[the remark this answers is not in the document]
dstathis: The protection on other calls are not necessary. I will remove them in a future commit.

src/charm.py
```diff
@@ -43,55 +46,94 @@ def new_func(self, *args, **kwargs):
...
+        for relation in self.model.relations["alerting"]:
+            self.update_alerting(relation)
+
+    def configure(self):
```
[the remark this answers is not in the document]
dstathis: I had it doing other things before and I would like to leave that option open. It is configuring the charm. If that happens to only have one step at the moment so be it.

src/charm.py
dstathis: I'm not sure I understand the comment here. update_alerting is used in multiple places and I don't want class code calling a callback.

See above comment about leader checking.

src/charm.py
dstathis: As this is a purely preference decision. I find code with fewer "not"s to be more satisfying.

src/charm.py
dstathis: You need to have some sort of alerting configured. Pagerduty is the only one currently supported as email would require me to write an smtp forwarder charm.

run_tests
```diff
@@ -12,7 +12,8 @@ else
     export PYTHONPATH="src:$PYTHONPATH"
 fi
 
-flake8
+flake8 || exit $?
```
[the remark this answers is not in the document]
dstathis: Good point

PR #5, opened 2020-12-10
a couple tests

dstathis:


tests/test_charm.py
```diff
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 
 import unittest
+import ops
```
justinmclark, line 5: Would it be better to do `from ops.model import BlockedStatus` rather than importing the whole framework?

justinmclark (approved):
One comment, but it looks good otherwise.

tests/test_charm.py
dstathis: Do you know if importing just BlockedStatus is more efficient?

dstathis: Interestingly it would be slightly more efficient here to import the whole module

https://stackoverflow.com/a/346967

Not that it really makes a difference

justinmclark: Interesting.

I also found this: https://softwareengineering.stackexchange.com/a/187471

So, the only real reason to do `from ops.model import BlockedStatus` seems to be stylistic.

run_tests
```diff
@@ -13,7 +13,7 @@ else
 fi
 
 flake8 || exit $?
-black --diff --check --skip-string-normalization src tests || exit $?
+black --diff --check src tests || exit $?
 which pytest > /dev/null
```
Abuelodelanada, line 17: Are we going to use pytest here?

Abuelodelanada (commented):
Looks good to me

run_tests
dstathis: It uses pytest if installed

Abuelodelanada: @dstathis I was wondering if shouldn't we run: `black -l 79 --diff --check src tests || exit $?` so we would comply with [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

dstathis: Flake8 runs on the previous line and will catch that

dstathis: @jlounder Do we have a Canonical standard for line length?

balbirthomas (approved):

PR #7, opened 2021-06-07
pebble conversion

sed-i:


src/provider.py
```diff
@@ -0,0 +1,43 @@
...
+import ops.model
+from ops.relation import ProviderBase
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+# TODO: name class after the relation?
+class AlertmanagerProvider(ProviderBase):
```
balbirthomas, line 15: It is not clear to me who are the intended users of `AlertmanagerProvider` ? Is it integrator charms ? Or is it Prometheus ? Some docstrings here may help.

src/provider.py
```diff
@@ -0,0 +1,43 @@
...
+        super().__init__(charm, self._provider_relation_name, service_name, version)
+        self.charm = charm
+        self._service_name = service_name
+
+        events = self.charm.on[self._provider_relation_name]
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+        self.framework.observe(events.relation_departed, self._on_relation_departed)
+        self.framework.observe(events.relation_broken, self._on_relation_broken)
+
+    def _on_relation_joined(self, event: ops.charm.RelationJoinedEvent):
```
balbirthomas, line 29: Same question regarding purpose of these relation handlers ?

tests/test_charm.py
```diff
@@ -17,14 +26,26 @@ def setUp(self):
         self.harness.set_leader(True)
         self.harness.update_config({"pagerduty_key": "123"})
 
+    @unittest.skip("")
     def test_config_changed(self):
+        def get_config():
+            pod_spec = self.harness.get_pod_spec()
+            config_yaml = pod_spec[0]["containers"][0]["volumeConfig"][0]["files"][0][
```
balbirthomas, line 33: Style comment : How about a one line utility method to make this `pod_spec[0]["containers"][0]["volumeConfig"][0]["files"][0] ...` more readable ?

tests/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # unless = None, args = lists
+        self.assertEqual([1, 2, 3], utils.append_unless(None, [1, 2], [3]))
+        self.assertEqual(None, utils.append_unless(None, None, [3]))
+
+    def test_md5(self):
+        self.assertEqual('50e1eeff19b2501c791612cd907fff7a',
+                         utils.md5('test string\nwith newline'))
+
+    def test_fetch_url(self):
```
balbirthomas, line 26: Aren't you effectively testing a third party library, given your `fetch_url` is a very lightweight wrapper around one of its methods. Is it worth this much effort :-) .

balbirthomas (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
+# Contributing to alertmanager-k8s
```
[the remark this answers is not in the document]
sed-i: Added

README.md
```diff
@@ -101,8 +51,22 @@ Install the development requirements
 
     pip install -r requirements-dev.txt
 
-## Testing
+Later on, upgrade packages as needed
```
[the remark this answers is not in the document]
sed-i: Removed

lib/charms/alertmanager-k8s/v1/consumer.py
```diff
@@ -0,0 +1,83 @@
...
+
+LIBID = "abcdef1234"  # Unique ID that refers to the library forever
+LIBAPI = 1    # Must match the major version in the import path.
+LIBPATCH = 0  # The current patch version. Must be updated when changing.
+
+logger = logging.getLogger(__name__)
+
+
+class AlertmanagerConsumer(ConsumerBase):
+    """A one sentence summary of the class.
```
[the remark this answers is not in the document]
sed-i: Updated

lib/charms/alertmanager-k8s/v1/consumer.py
```diff
@@ -0,0 +1,83 @@
...
+    Attributes:
+            charm (CharmBase): consumer charm
+    """
+    _stored: StoredState
+
+    def __init__(self, charm: ops.charm.CharmBase, relation_name, consumes, multi=False):
+        super().__init__(charm, relation_name, consumes, multi)
+        self.charm = charm
+        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
+        self._provider_relation_name = "alerting"  # from alertmanager's metadata.yaml
```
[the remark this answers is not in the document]
sed-i: It is indeed unused. Commented out for now.

lib/charms/alertmanager-k8s/v1/consumer.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_changed,
+                               self._on_relation_changed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_departed,
+                               self._on_relation_departed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_broken,
+                               self._on_relation_broken)
+
+        self._stored.set_default(alertmanagers={})
```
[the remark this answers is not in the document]
sed-i: When units come and go, I need to know which address belongs to who when I pop it out in `on_relation_departed`, where I only have the originating unit.

lib/charms/alertmanager-k8s/v1/consumer.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _on_relation_changed(self, event: ops.charm.RelationChangedEvent):
```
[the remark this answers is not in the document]
sed-i: Added

lib/charms/alertmanager-k8s/v1/consumer.py
```diff
@@ -0,0 +1,83 @@
...
+                self._stored.alertmanagers[event.unit.name] = address
+
+        if self.charm.unit.is_leader():
+            # forward list of IP addresses # TODO use app data bag instead
+            self.charm._stored.alertmanagers = sorted(list(self._stored.alertmanagers.values()))
+
+            # TODO figure out how to emit instead
+            self.charm._on_alertmanager_available(event)
+
+    def _on_relation_departed(self, event: ops.charm.RelationDepartedEvent):
```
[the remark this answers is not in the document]
sed-i: 1. Docstring Added.
2. Will experiment with app data shortly, thanks for pointing out.

metadata.yaml
```diff
@@ -1,23 +1,44 @@
-name: alertmanager
-summary: Alert management for charmed LMA stack
+name: alertmanager-k8s
+
+summary: |
+  Kubernetes charm for Alertmanager.
+
 maintainers:
   - Dylan Stephano-Shachter <dylan.stephano-shachter@canonical.com>
```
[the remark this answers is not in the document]
sed-i: Removed `maintainers` entry entirely

src/provider.py
sed-i: afaiu, the provider is _always_ used _only_ by the "main" charm, i.e. Alertmanager.

src/provider.py
sed-i: Docstring added

tests/test_charm.py
sed-i: This is an old test that will soon be rewritten. I kept it for now for reference.

tests/test_utils.py
sed-i: The purpose of the test is to test the interface for `fetch_url`, and to that end I could have used `with patch(...)` instead. Thanks for pointing out.

lib/charms/alertmanager-k8s/v1/consumer.py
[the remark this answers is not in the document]
sed-i: @jlounder, please advise:
iiuc, "public address" is relation-bound, which means that if we have 2 units of prometheus and 1 unit of alertmanager then alertmanager may have a different address for each prometheus.
If this is the case, then managing this in app data would be much more involved compared to the local nature of `_stored`.

lib/charms/alertmanager_k8s/v1/consumer.py
```diff
@@ -0,0 +1,92 @@
...
+        self._stored.set_default(alertmanagers={})
+
+    def _on_relation_changed(self, event: ops.charm.RelationChangedEvent):
+        """This hook stores locally the address of the newly-joined alertmanager.
+        This is needed for consumers such as prometheus, which should be aware of all alertmanager
+        instances.
+        """
+        if event.unit:  # event.unit may be `None` in the case of app data change
+            # Save locally the public IP address of the alertmanager unit
+            if address := event.relation.data[event.unit].get("public_address"):
```
rbarry82, line 62: Walrus won't work if we're targeting Py3.5

sed-i: I thought we said that k8s charms, running on 20.04, can assume py3.8

src/provider.py
balbirthomas: Yes Indeed. Provider is always your charm, in this case Alertmanager.

tests/test_utils.py
balbirthomas: `with patch(...)` sounds simpler.

metadata.yaml
```diff
@@ -1,23 +1,42 @@
...
+
 provides:
   alerting:
     interface: alertmanager
+    # assumed network type: private
+
 peers:
-  alertmanager:
-    interface: alertmanager
+  replicas:
```
mmanciop, line 36: I like!

metadata.yaml
```diff
@@ -1,23 +1,42 @@
...
+
+# oci-image resources for each container defined above
+resources:
+  alertmanager-image:
+    type: oci-image
+    description: OCI image for alertmanager
+
 provides:
   alerting:
     interface: alertmanager
```
mmanciop, line 32: I think this should be “alertmanager-ingestion”, as opposed to the APIs that Karma will use

mmanciop (changes requested):
I think we need to make the interface name a bit more specific because Alertmanager will integrate in different ways with Prometheus and Karma. The rest LGTM, although on mobile I could not really review the README

run_tests
```diff
@@ -1,22 +1,19 @@
-#!/bin/bash
-# Copyright 2020 dylan
+#!/bin/sh -e
```
dstathis, line 1: Whatever you prefer here but I would maybe consider `set -e` rather than changing the shell

metadata.yaml
sed-i: how would you name it in prometheus `requires`? also `alertmanager-ingestion`?

src/charm.py
```diff
@@ -1,217 +1,416 @@
...
+    _storage_path = "/alertmanager"
+
+    _stored = StoredState()
+
     def __init__(self, *args):
-        log.debug("Initializing charm.")
         super().__init__(*args)
-        self.framework.observe(self.on.config_changed, self.on_config_changed)
+        self.container = self.unit.get_container(self._container_name)
+        self.provider = AlertmanagerProvider(self, self._service_name, "0.0.1")
```
dstathis, line 79: I think the version string here is supposed to be the version of alertmanager. (probably should be higher than 0.0.1)

src/charm.py
```diff
@@ -1,217 +1,416 @@
...
+    def _fetch_private_address(self) -> Optional[str]:
+        """Fetch private address from unit's peer relation data bucket.
+        """
+        return self.peer_relation.data[self.unit].get("private_address")
+
+    def _alertmanager_layer(self) -> Dict[str, Any]:
+        """Returns Pebble configuration layer for alertmanager.
+        """
+
+        def _command():
```
dstathis, line 153: Why is this defined inside `_alertmanager_layer` rather than being its own method?

src/charm.py
```diff
@@ -1,217 +1,416 @@
...
+    _service_name: str = "alertmanager"  # chosen arbitrarily to match charm name
+    _peer_relation_name: str = "replicas"  # must match metadata.yaml peer role name
+    _api_port: int = 9093  # port to listen on for the web interface and API
+    _ha_port: int = 9094  # port for HA-communication between multiple instances of alertmanager
+
+    # path, inside the workload container, to the alertmanager configuration file
+    _config_path = "/etc/alertmanager/alertmanager.yml"
+
+    # path, inside the workload container for alertmanager logs, e.g. 'nflogs', 'silences'.
+    _storage_path = "/alertmanager"
```
dstathis, line 72: I think the more common python pattern would be to make these globals. This works too though

src/charm.py
```diff
@@ -1,217 +1,416 @@
...
+            # This reduces down-time compared to restarting the service.
+            api = AlertmanagerAPIClient(self._fetch_private_address(), self._api_port)
+            if not api.reload():
+                logger.warning("config reload via HTTP POST failed")
+                if restart_on_failure:
+                    self._restart_service()
+
+        return is_changed
+
+    def _common_exit_hook(self) -> bool:
```
dstathis, line 340: This name is a bit confusing. There seems to be a lot more logic here than just exiting the hook

dstathis (changes requested):

run_tests
```diff
@@ -1,22 +1,19 @@
...
-flake8 || exit $?
-black --diff --check src tests || exit $?
-which pytest > /dev/null
-if [ $? -eq 0 ]; then
-    python3 -m pytest -v "$@"
-else
-    python3 -m unittest -v "$@"
-fi
+flake8
+# black --quiet --diff --check src tests || exit $?
```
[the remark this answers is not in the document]
sed-i: black is back

config.yaml
```diff
@@ -1,17 +1,9 @@
...
 #
 # This is only an example, and you should edit to suit your needs.
 # If you don't need config, you can remove the file entirely.
+
 options:
-  alertmanager_image_path:
-    default: 'quay.io/prometheus/alertmanager'
-    description: The docker image to use
-    type: string
-  port:
```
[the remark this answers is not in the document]
sed-i: Removal of `port` confirmed by @jlounder and @mmanciop (random port can still be scanned so no added security).

requirements.txt
```diff
@@ -1,3 +1,5 @@
-# ops 1.0.0 can not get addresses on kubernetes due
-# to the fact that the interfaces have no names
-ops>=1.0.1
+#ops >= 1.2.0
```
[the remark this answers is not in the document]
sed-i: Removed

run_tests
sed-i: I think I used [this](https://github.com/jnsgruk/charm-kubernetes-dashboard/blob/master/run_tests) as a reference at the time. `/bin/sh` might be more portable if `bash` functionality is not need. Either way, the preferred shebang for bash seems to be `#!/usr/bin/env bash`.

src/charm.py
sed-i: The docstring of `ProviderBase` suggests that you are right:
```python
            It is important that the version is obtained by actually
            querying the deployed application rather than being written
            into the code.
```
@balbirthomas could you please confirm? i.e. should the version string be based on the output of `/bin/alertmanager --version` in the underlying oci-image?

src/charm.py
sed-i: For encapsulation/hiding reasons: it is only needed there and should probably be never used in other context.

src/charm.py
sed-i: Yeah I had it as globals at first, but module-wide variables for this seem like an anti-pattern to me tbh.
1. I needed them in the provider and tests, so `self.charm._api_port` and `self.harness.charm._api_port` seem more correct than `from charm import API_PORT`.
2. Technically, a single `charm.py` could have two different charm classes. Not that we'll ever do it.

src/charm.py
sed-i: Please do suggest a better name. My intention was not "exiting the hook" but rather a common hook through which everyone is exiting, e.g. `_common_gate`, `_one_hook_to_rule_them_all` etc.

dstathis: I have it as just _configure but it's up to you. I don't even really see this as a problem. I just thought I would mention that one might think this is just exit logic.

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,146 @@
...
+    def _on_relation_joined(self, event: ops.charm.RelationJoinedEvent):
+        """This hook stores the public address of the newly-joined "alerting" relation in the
+        corresponding data bag.
+        This is needed for consumers such as prometheus, which should be aware of all alertmanager
+        instances.
+        """
+        # "ingress-address" is auto-populated incorrectly so rolling my own, "public_address"
+        event.relation.data[self.charm.unit]["public_address"] = "{}:{}".format(
+            self.model.get_binding(event.relation).network.bind_address, self._public_api_port
+        )
```
dstathis, line 146: Rather than pulling _public_api_port from the charm, the library should implement a "set_public_address" method that the charm calls on a relation_joined event. This is to make the library more portable is someone else wants to implement an alertmanager charm. They can reuse the library without needing to have specific private members of their class.

src/charm.py
dstathis: The library should not be calling in to charm code. Even if it is just a constant. I made another comment about this.

src/charm.py
dstathis: This feels very counter to standard practice. _command is not really hidden from anything inside there and its weird to have the method redefined on every call to _alertmanager_layer (and maybe inefficient but I don't know)

lib/charms/alertmanager_k8s/v0/alertmanager.py
sed-i: Good point! Added.

src/charm.py
balbirthomas: The version string being provided by the provider should not be hardcoded but rather querried from the actual altermanager running. This is because the system adminstrator may choose to deploy a different image of the alertmanager OCI container which can have a different version. Hopefully this clarifies the issue.

src/charm.py
sed-i: Is `from charm import API_PORT` not "charm code"?

src/charm.py
sed-i: Nested functions are very common in factories and closures.
Encapsulation is also common.
Another way to encapsulate the same would be e.g. `class AlertmanagerPebbleLayer`.

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,146 @@
...
+
+                # inform consumer about the change
+                # manually emitting the "available" event until scale-up/down "ready" flag is fixed
+                self.on.available.emit()
+
+    def get_cluster_info(self) -> List[str]:
+        """Returns a list of ip addresses of all the alertmanager units"""
+        return sorted(list(self._stored.alertmanagers.values()))
+
+    def _on_relation_departed(self, event: ops.charm.RelationDepartedEvent):
```
balbirthomas, line 90: I do not see why you need to be reimplementing all these event handlers and then raising `ProviderAvailable` . Same question with regard to the relation broken handler below.

balbirthomas (changes requested):

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        if event.unit:  # event.unit may be `None` in the case of app data change
+            # Save locally the public IP address of the alertmanager unit
+            if address := event.relation.data[event.unit].get("public_address"):
+                # TODO consider storing in unit data instead of StoredState
+                self._stored.alertmanagers[event.unit.name] = address
+
+                # inform consumer about the change
+                # manually emitting the "available" event until scale-up/down "ready" flag is fixed
+                self.on.available.emit()
```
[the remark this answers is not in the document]
sed-i: This is a very good argument in favor of new custom event (e.g. `ClusterChangedEvent`), because if I try to stick to OF events then the consumer charms would have to register available, invalid, etc.

src/charm.py
dstathis: okay

lib/charms/alertmanager_k8s/v0/alertmanager.py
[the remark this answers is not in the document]
sed-i: Replaced with a custom event.

lib/charms/alertmanager_k8s/v0/alertmanager.py
sed-i: Replaced with a custom event.

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,175 @@
...
+    _stored: StoredState
+    cluster_changed = EventSource(ClusterChanged)
+
+    def __init__(self, charm: CharmBase, relation_name: str, consumes: dict, multi: bool = False):
+        super().__init__(charm, relation_name, consumes, multi)
+        self.charm = charm
+        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
+
+        self.framework.observe(
+            self.charm.on[self._consumer_relation_name].relation_changed, self._on_relation_changed
```
balbirthomas, line 79: The structure of this library is certainly improved however I am still concerned that on a relation changed event this library can emit a `ClusterChanged` event while `ConsumerBase` emits a `ProviderInvalid`. Do you see this being a possibility ? If this does happen wouldn't this result in Prometheus adding an alertmanager into its config that is and "invalid" alertmanager ?

sed-i: `ClusterChanged` event while `ProviderInvalid` is probably a possible scenario, but I think there is no wrong with it: 
1. alertmanager having an IP address is independent of the provider being `ready` or `unready`. Moreover, taking into account `ConsumerEvents` would not be easy, and we also do not have a pattern for it yet, afaik.
2. if an alertmanager unit joins but the provider did not set `ready` yet, it's ok for a consumer charm to have the ip address.
3. if an alertmanager unit leaves and the provider is unready, it makes sense for the consumer to remove it.

balbirthomas: In invalid provider has to do with a version or type incompatibility with the consumer, in dependent of provider is ready or not. If you agree `ClusterChanged` can be raised when provider is invalid this means that Prometheus config will have alertmanager that it is version incompatible with. Is this a bug ?

sed-i: oh ok I see, so `ProviderInvalid` would be emitted once and that's it right?
In that case:
1. "version mismatch", as far as alertmanager and say prometheus are concerned, is probably an "unsupported api" error, e.g. prometheus tries to access `/api/v2`, which doesn't exist for the version of alertmanager in use. But this is hypothetical and probably the responsibility of prometheus: prometheus should query alertmanager's `/api/status/version` and use the correct alertmanager api endpoints.
2. Taking into account ProviderInvalid would not be easy and we do not have a pattern for it yet, afaik.

balbirthomas: You are saying that the Prometheus charm should be responsible for ensuring it can deal with an incompatible version of an Alertmanager charm, despite the fact that

1. The Prometheus charm (the consumer) specifies the minimum Alertmanager version using a consumers clause.
2. The Provider Consumer library raises a ProviderInvalid event.

I do not see what then is the point of the Provider Consumer library. Do you ? 

With regard to ready and unready. Can't it happen that the Provider (Alertmanager) is ready but still incompatible with Prometheus ?

dstathis (approved):

lib/charms/alertmanager_k8s/v0/alertmanager.py
sed-i: Captured in #9, will address in next PR.

balbirthomas (approved):

PR #17, opened 2021-07-13
fix docstrings

sed-i:


balbirthomas (approved):

src/charm.py
```diff
@@ -66,12 +66,12 @@ def version(self) -> Optional[str]:
 
 
 class AlertmanagerCharm(CharmBase):
-    _container_name: str = "alertmanager"  # automatically determined from charm name
+    _container_name = "alertmanager"  # automatically determined from charm name
```
jnsgruk, line 69: you could probably shorten this: `_conainer_name = _layer_name = _service_name = "alertmanager"`

src/charm.py
```diff
@@ -66,12 +66,12 @@ def version(self) -> Optional[str]:
...
     _layer_name = "alertmanager"  # layer label argument for container.add_layer
-    _service_name: str = "alertmanager"  # chosen arbitrarily to match charm name
-    _peer_relation_name: str = "replicas"  # must match metadata.yaml peer role name
-    _api_port: int = 9093  # port to listen on for the web interface and API
-    _ha_port: int = 9094  # port for HA-communication between multiple instances of alertmanager
+    _service_name = "alertmanager"  # chosen arbitrarily to match charm name
+    _peer_relation_name = "replicas"  # must match metadata.yaml peer role name
+    _api_port = 9093  # port to listen on for the web interface and API
+    _ha_port = 9094  # port for HA-communication between multiple instances of alertmanager
```
jnsgruk, line 74: If these are not configurable, perhaps better as constants for now?

src/charm.py
```diff
@@ -66,12 +66,12 @@ def version(self) -> Optional[str]:
...
 class AlertmanagerCharm(CharmBase):
-    _container_name: str = "alertmanager"  # automatically determined from charm name
+    _container_name = "alertmanager"  # automatically determined from charm name
     _layer_name = "alertmanager"  # layer label argument for container.add_layer
-    _service_name: str = "alertmanager"  # chosen arbitrarily to match charm name
-    _peer_relation_name: str = "replicas"  # must match metadata.yaml peer role name
-    _api_port: int = 9093  # port to listen on for the web interface and API
-    _ha_port: int = 9094  # port for HA-communication between multiple instances of alertmanager
+    _service_name = "alertmanager"  # chosen arbitrarily to match charm name
+    _peer_relation_name = "replicas"  # must match metadata.yaml peer role name
```
jnsgruk, line 72: You could get this dynamically using the [`CharmMeta`](https://ops.readthedocs.io/en/latest/#ops.charm.CharmMeta) class

jnsgruk: Same for `_storage_path` below

jnsgruk (commented):
Nitpicks only because this PR slid into my notifications and caught my eye :)

src/charm.py
sed-i: - peers: do you mean `next(iter(self.meta.peers.keys()))` under the assumption there is only one?
- storage: in metadata.yaml, I have `storage:  data: ...`, where "data" is arbitrary, so I'd have to hard code something anyway: `self.meta.storages['data'].location`, unless I assume there is only one. I actually do prefer hard-coding "data" over hard-coding the location itself, but just wanted to make sure what you meant.

src/charm.py
sed-i: Done

src/charm.py
sed-i: do you mean literals? I made this to avoid duplicating literals

sed-i: unless you meant module-wide variables in all-caps? If this is debatable then I'd argue for scope correctness.

src/charm.py
jnsgruk: On peers: yes - it might be prudent to check there is indeed only one peer relation

On storage - indeed, that is what I meant :)

sed-i: Done

src/charm.py
```diff
@@ -66,25 +66,26 @@ def version(self) -> Optional[str]:
...
-    _service_name: str = "alertmanager"  # chosen arbitrarily to match charm name
-    _peer_relation_name: str = "replicas"  # must match metadata.yaml peer role name
-    _api_port: int = 9093  # port to listen on for the web interface and API
-    _ha_port: int = 9094  # port for HA-communication between multiple instances of alertmanager
+    # Container name is automatically determined from charm name
+    # Layer name is used for the layer label argument in container.add_layer
+    # Service name is chosen arbitrarily to match charm name
+    _container_name = _layer_name = _service_name = "alertmanager"
+    _peer_relation_name = "replicas"  # must match metadata.yaml peer role name
+    _api_port = 9093  # port to listen on for the web interface and API
```
rbarry82, line 74: Why are these not in `config.yaml`?

sed-i: The port number used to be, but since it is internal to the container it shouldn't matter for the user.

rbarry82: I don’t agree with this assumption. Even if administrators may not change this often, once the nginx-ingress operator supports [arbitrary TCP/UCP remapping](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/exposing-tcp-udp-services.md), having this information at least visible (even if it’s not configurable) will be useful for administrators via something like `service-port` in the [existing charm](https://charmhub.io/nginx-ingress-integrator) if they want to expose it. 

This is a small change now (add it back to `config.yaml` and use `self.model.config[“port”]` instead) to future proof and save yourself headache later

src/charm.py
dstathis: It's definitely more common python practice to put them at the module level. The caps, however, I agree is unnecessary.

dstathis (approved):

src/charm.py
rbarry82: Is there another push which changes it here?

If not, but can get this dynamically via `next(iter(self.meta.peers)).name` or `list(self.meta.peers.keys())[0].name` or other. I don't think hardcoding it bad, but having a `# must match metadata.yaml peer role name` as a comment may make readers wonder "why don't you just fetch it from there?"

One or the other (hardcoding without the comment) may be better. Similarly, just above this, removing "arbitrarily" is probably a good idea, since that makes it sound as if there was not a design decision behind this, and there was -- consistency. `# Service name matches charm name for consistency` reads better.

Really a nit, though

src/charm.py
sed-i: Captured in https://github.com/canonical/alertmanager-operator/issues/19

src/charm.py
sed-i: There was a commit with next(iter(...)) but I reverted it after realizing more than one peer relation is possible.

PR #18, opened 2021-07-14
add relation to karma

sed-i:
this adds support for relating to karma in the exact same way [the proxy](https://github.com/sed-i/charm-alertmanager-karma-proxy) does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ Currently, supported relations are:
   - [Prometheus](https://github.com/canonical/prometheus-operator), which forwards alerts to
     Alertmanager over the `alertmanager_dispatch` interface. Set up with: 
     `juju add-relation alertmanager-k8s prometheus-k8s`.
+  - [Karma](https://github.com/canonical/karma-operator), which displays alerts from all related alertmanager instances 
+    in a nice dashboard.
+    Set up with `juju add-relation alertmanager-k8s karma-k8s`.
 
 
 ## OCI Images
--- a/README.md
+++ b/README.md
@@ -45,10 +45,10 @@ address of the Alertmanager unit, which is determinable with a `juju status` com
 
 Currently, supported relations are:
   - [Prometheus](https://github.com/canonical/prometheus-operator), which forwards alerts to
-    Alertmanager over the `alertmanager_dispatch` interface. Set up with: 
-    `juju add-relation alertmanager-k8s prometheus-k8s`.
+    Alertmanager over the `alertmanager_dispatch` interface.
+    Set up with `juju add-relation alertmanager-k8s prometheus-k8s`.
   - [Karma](https://github.com/canonical/karma-operator), which displays alerts from all related alertmanager instances 
-    in a nice dashboard.
+    over the `karma_dashboard` interface.
     Set up with `juju add-relation alertmanager-k8s karma-k8s`.
 
 
--- a/lib/charms/karma_k8s/v0/karma.py
+++ b/lib/charms/karma_k8s/v0/karma.py
@@ -14,7 +14,7 @@ from ops.charm import RelationJoinedEvent, RelationDepartedEvent
 from ops.relation import ConsumerBase, ProviderBase
 from ops.framework import StoredState
 
-from typing import List, Dict, Optional, Iterable
+from typing import List, Dict, Optional
 
 # The unique Charmhub library identifier, never change it
 LIBID = "abcdef1234"
@@ -36,12 +36,17 @@ class KarmaAlertmanagerConfig:
     def is_valid(config: Dict[str, str]):
         return all(key in config for key in KarmaAlertmanagerConfig.required_fields)
 
+    @staticmethod
+    def from_dict(data: Dict[str, str]) -> Dict[str, str]:
+        config = {k: data[k] for k in data if k in KarmaAlertmanagerConfig.required_fields}
+        return config if KarmaAlertmanagerConfig.is_valid(config) else {}
 
-def dict_subset(data: dict, key_subset: Iterable):
-    return {k: v for k, v in data.items() if k in key_subset}
+    @staticmethod
+    def build(name: str, uri: str) -> Dict[str, str]:
+        return KarmaAlertmanagerConfig.from_dict({"name": name, "uri": uri})
 
 
-class GenericEvent(EventBase):
+class KarmaAlertmanagerConfigChanged(EventBase):
     def __init__(self, handle, data=None):
         super().__init__(handle)
         self.data = data
@@ -53,10 +58,6 @@ class GenericEvent(EventBase):
     def restore(self, snapshot):
         """Restore relation data."""
         self.data = snapshot["data"]
-
-
-class KarmaAlertmanagerConfigChanged(GenericEvent):
-    pass
 
 
 class KarmaProviderEvents(ObjectEvents):
@@ -109,7 +110,7 @@ class KarmaProvider(ProviderBase):
 
     Arguments:
             charm (CharmBase): consumer charm
-            relation_name (str): from consumer's metadata.yaml
+            name (str): relation name from consumer's metadata.yaml
             service_name (str): service name (must be consistent the consumer)
             version (str): semver-compatible version string
 
@@ -120,13 +121,12 @@ class KarmaProvider(ProviderBase):
     on = KarmaProviderEvents()
     _stored = StoredState()
 
-    def __init__(self, charm, relation_name: str, service_name: str, version: str = None):
-        super().__init__(charm, relation_name, service_name, version)
+    def __init__(self, charm, name: str, service_name: str, version: str = None):
+        super().__init__(charm, name, service_name, version)
         self.charm = charm
-        self._relation_name = relation_name
         self._service_name = service_name
 
-        events = self.charm.on[self._relation_name]
+        events = self.charm.on[self.name]
         self.framework.observe(events.relation_joined, self._on_relation_joined)
         self.framework.observe(events.relation_changed, self._on_relation_changed)
         self.framework.observe(events.relation_departed, self._on_relation_departed)
@@ -134,10 +134,8 @@ class KarmaProvider(ProviderBase):
 
     def get_alertmanager_servers(self) -> List[Dict[str, str]]:
         alertmanager_ips = []
-        logger.info("get_alertmanager_servers")
 
-        for relation in self.charm.model.relations[self._relation_name]:
-            logger.info("relation.data = %s", relation.data)
+        for relation in self.charm.model.relations[self.name]:
             if relation.id not in self._stored.active_relations:
                 # relation id is not present in the set of active relations
                 # this probably means that RelationBroken did not exit yet (was recently removed)
@@ -150,8 +148,8 @@ class KarmaProvider(ProviderBase):
                     data = relation.data[key]
                     break
             if data:
-                config = dict_subset(data, KarmaAlertmanagerConfig.required_fields)
-                if KarmaAlertmanagerConfig.is_valid(config) and config not in alertmanager_ips:
+                config = KarmaAlertmanagerConfig.from_dict(data)
+                if config and config not in alertmanager_ips:
                     alertmanager_ips.append(config)
             else:
                 logger.warning("no related units in relation dict")
@@ -218,7 +216,7 @@ class KarmaConsumer(ConsumerBase):
 
     Arguments:
             charm (CharmBase): consumer charm
-            relation_name (str): from consumer's metadata.yaml
+            name (str): from consumer's metadata.yaml
             consumes (dict): provider specifications
             multi (bool): multiple relations flag
 
@@ -228,21 +226,15 @@ class KarmaConsumer(ConsumerBase):
 
     _stored = StoredState()
 
-    def __init__(self, charm, relation_name: str, consumes: dict, multi: bool = False):
-        super().__init__(charm, relation_name, consumes, multi)
+    def __init__(self, charm, name: str, consumes: dict, multi: bool = False):
+        super().__init__(charm, name, consumes, multi)
         self.charm = charm
-        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
         self._stored.set_default(config={})
-
-        events = self.charm.on[self._consumer_relation_name]
-
+        events = self.charm.on[self.name]
         self.framework.observe(events.relation_joined, self._on_relation_joined)
 
     def _on_relation_joined(self, event: RelationJoinedEvent):
-        if not self.model.unit.is_leader():
-            return
-
-        # update app data bag
+        # update relation data bag
         event.relation.data[self.charm.unit].update(self._stored.config)
 
     @property
@@ -254,16 +246,15 @@ class KarmaConsumer(ConsumerBase):
         return self._stored.config.get("uri", None)
 
     def set_config(self, name: str, uri: str) -> bool:
-        config = {"name": name, "uri": uri}
-        if not KarmaAlertmanagerConfig.is_valid(config):
+        if not (config := KarmaAlertmanagerConfig.build(name, uri)):
             logger.warning("Invalid config: {%s, %s}", name, uri)
             return False
 
         self._stored.config.update(config)
         logger.info("stored config: %s", self._stored.config)
 
-        if self.model.unit.is_leader() and self._consumer_relation_name in self.charm.model.relations:
-            for relation in self.charm.model.relations[self._consumer_relation_name]:
+        if self.model.unit.is_leader() and self.name in self.charm.model.relations:
+            for relation in self.charm.model.relations[self.name]:
                 relation.data[self.charm.unit].update(self._stored.config)
 
         return True
--- a/lib/charms/karma_k8s/v0/karma.py
+++ b/lib/charms/karma_k8s/v0/karma.py
@@ -86,7 +86,7 @@ class KarmaProvider(ProviderBase):
 
     ```python
     self.provider = KarmaProvider(
-        self, "karma-dashboard", "karma", "0.0.1"
+        self, "karma-dashboard", "karma", "0.86"
     )
     ```
 
--- a/lib/charms/karma_k8s/v0/karma.py
+++ b/lib/charms/karma_k8s/v0/karma.py
@@ -1,0 +1,269 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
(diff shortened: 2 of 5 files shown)


lib/charms/karma_k8s/v0/karma.py
```diff
@@ -86,7 +86,7 @@ class KarmaProvider(ProviderBase):
 
     ```python
     self.provider = KarmaProvider(
-        self, "karma-dashboard", "karma", "0.0.1"
+        self, "karma-dashboard", "karma", "0.86"
```
rbarry82, line 89: May want to make it `>=` or `<=` just for safety

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "abcdef1234"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 2
```
rbarry82, line 27: Should this be `2`?

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+logger = logging.getLogger(__name__)
+
+
+class KarmaAlertmanagerConfig:
+    required_fields = ("name", "uri")
+
+    @staticmethod
+    def is_valid(config: Dict[str, str]):
+        return all(key in config for key in KarmaAlertmanagerConfig.required_fields)
+
```
rbarry82, line 38: This could easily be a simple method instead of a class with a single static method

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+
+def dict_subset(data: dict, key_subset: Iterable):
```
rbarry82, line 40: Only used once. Feel free to keep, but could be a lambda or just a comprehension instead of making people read it look up a single use function

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+    return {k: v for k, v in data.items() if k in key_subset}
+
+
+class GenericEvent(EventBase):
```
rbarry82, line 44: Only used once. WHy is `KarmaAlertManagerConfigChanged` defined here instead of `GenericEvent`? Do we plan to add more events in this library?

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+            charm (CharmBase): consumer charm
+    """
+
+    on = KarmaProviderEvents()
+    _stored = StoredState()
+
+    def __init__(self, charm, relation_name: str, service_name: str, version: str = None):
+        super().__init__(charm, relation_name, service_name, version)
+        self.charm = charm
+        self._relation_name = relation_name
```
rbarry82, line 126: Already available as `self.name` from `super().name`

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+
+        events = self.charm.on[self._relation_name]
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+        self.framework.observe(events.relation_changed, self._on_relation_changed)
+        self.framework.observe(events.relation_departed, self._on_relation_departed)
+        self._stored.set_default(active_relations=set())
+
+    def get_alertmanager_servers(self) -> List[Dict[str, str]]:
+        alertmanager_ips = []
+        logger.info("get_alertmanager_servers")
```
rbarry82, line 137: `logger.debug`

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+
+        for relation in self.charm.model.relations[self._relation_name]:
+            logger.info("relation.data = %s", relation.data)
```
rbarry82, line 140: Again. Probably a lot. May as well change them before merging

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+    Attributes:
+            charm (CharmBase): consumer charm
+    """
+
+    _stored = StoredState()
+
+    def __init__(self, charm, relation_name: str, consumes: dict, multi: bool = False):
+        super().__init__(charm, relation_name, consumes, multi)
+        self.charm = charm
+        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
```
rbarry82, line 234: Is it from the metadata? I'm not sure how you're calling this. It may be useful to give an example, because otherwise it's just a `str`

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,269 @@
...
+        events = self.charm.on[self._consumer_relation_name]
+
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+
+    def _on_relation_joined(self, event: RelationJoinedEvent):
+        if not self.model.unit.is_leader():
+            return
+
+        # update app data bag
+        event.relation.data[self.charm.unit].update(self._stored.config)
```
rbarry82, line 246: Since this is a unit data bag, you don't need the guard above

rbarry82 (approved):
LGTM in general. The separate unit databags and collating them all later is a good pattern we could use across all of our "HA is shared-nothing redundancy" applications

lib/charms/karma_k8s/v0/karma.py
sed-i: the ProviderBase doc string implies it should be a fixed number:

```
        version: a string providing version of service provided by the
            charm. [...]
            It is important that the version is obtained by actually
            querrying the deployed application rather than being written
            into the code. This is because the same charm may be used
            to deploy different versions of a service (application).
```

lib/charms/karma_k8s/v0/karma.py
sed-i: Was handed over to me like that, I think. Should probably be kept monotonic and rising even if not published yet.
Unless there is a reason to revert back to 1?

lib/charms/karma_k8s/v0/karma.py
sed-i: Agreed. But I think this is going to [expand a bit](https://github.com/prymitive/karma/blob/main/docs/CONFIGURATION.md#alertmanagers).
As part of that I was planning on replacing `set_config(self, name: str, uri: str)` with `set_config(self, config: KarmaAlertmanagerConfig)` in the near future.

lib/charms/karma_k8s/v0/karma.py
sed-i: Good point. Removed.
BTW: `PEP 8: E731 do not assign a lambda expression, use a def`

lib/charms/karma_k8s/v0/karma.py
sed-i: The opposite: there used to be more events. Removed.

lib/charms/karma_k8s/v0/karma.py
sed-i: Thanks! Updated.

lib/charms/karma_k8s/v0/karma.py
sed-i: Good catch. Removed entirely.

lib/charms/karma_k8s/v0/karma.py
sed-i: Good catch. Removed entirely.

lib/charms/karma_k8s/v0/karma.py
sed-i: Simply replaced with `self.name` :)

lib/charms/karma_k8s/v0/karma.py
sed-i: Good catch! Used to be app data bag. Updated comment as well.

lib/charms/karma_k8s/v0/karma.py
mmanciop: I don't think we should revert it. Although blast radius of this is probably restricted to Xav :-)

README.md
```diff
@@ -47,6 +47,9 @@ Currently, supported relations are:
   - [Prometheus](https://github.com/canonical/prometheus-operator), which forwards alerts to
     Alertmanager over the `alertmanager_dispatch` interface. Set up with: 
     `juju add-relation alertmanager-k8s prometheus-k8s`.
+  - [Karma](https://github.com/canonical/karma-operator), which displays alerts from all related alertmanager instances 
+    in a nice dashboard.
```
mmanciop, line 51:
```
in [Karma](https://github.com/prymitive/karma), which is charmed by the [Karma operator](https://charmhub.io/karma-k8s).
```

sed-i: docstring improved to match the prometheus line above.

lib/charms/karma_k8s/v0/karma.py
rbarry82: It does support comparators, though.

sed-i: semver does, but from OF POV, the version the provider reports needs to be the version of the service/executable, i.e. the output of the `karma --version`. tbh I do not understand why would it be sensible to construct a provider with non-fixed version.

rbarry82: OF does not care. This is used for checking only whether the provider and consumer are compatible. Updating this string for every revision and finding that a consumer which specifies `0.87` is not actually compatible even though the API has not changed at all is going to be an annoying, frustrating experience. The version matters for this reason. The goal of `LIBAPI` is to say “hey, the library API is different”. The goal of the version in `ops.relation` is to say “ok, the library API is the same, but the workload container changed its configuration format, and the data we give you back will cause it to fail to start”. 

The exact version is not important. The operator framework does not (and arguably cannot) verify this, for the same reason as it does not support a “hot reload”. The flags to check it from the binary may be different, not exposed at all, etc. In addition, this is not exposed in `juju status`, so it has no meaning to users or administrators. 

https://github.com/canonical/operator/blob/92c7922e451b14184ceb9e7170d1564306fa1046/ops/relation.py#L482

I’d ask myself “do I want to submit a new PR here and force external consumers of the library to update theirs every single time there is a version bump, even if it doesn’t chance the `Provider|Consumer` API and the workload API|configuration syntax hasn’t changed at all”?

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,260 @@
...
+        self.charm = charm
+        self._service_name = service_name
+
+        events = self.charm.on[self.name]
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+        self.framework.observe(events.relation_changed, self._on_relation_changed)
+        self.framework.observe(events.relation_departed, self._on_relation_departed)
+        self._stored.set_default(active_relations=set())
+
+    def get_alertmanager_servers(self) -> List[Dict[str, str]]:
```
balbirthomas, line 135: There is surprisingly no docstring for a public method. Knowing the return value is a "list of dictionaries with both keys and values as strings" still requires the client programmer to read the code or guess that the key **may** be a name and the value **may** be an address.

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,260 @@
...
+
+    ```python
+    self.karma_lib = KarmaConsumer(
+        self,
+        "karma-dashboard",
+        consumes={"karma": ">=0.0.1"},
+    )
+    ```
+
+    The consumer charm is expected to set config via the consumer library, for example in config-changed:
```
balbirthomas, line 208: Perhaps direct the charm writer to where the structure of this "config" is specified.

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @property
+    def config_valid(self):
+        return KarmaAlertmanagerConfig.is_valid(self._stored.config)
+
+    @property
+    def ip_address(self) -> Optional[str]:
+        return self._stored.config.get("uri", None)
+
+    def set_config(self, name: str, uri: str) -> bool:
```
balbirthomas, line 248: Public method with no docstring.

balbirthomas (approved):

lib/charms/karma_k8s/v0/karma.py
sed-i: I don't see where the problem is: provider says "I am 0.86" (currently hard-coded, but in the future will be taken from container app) and consumer needs ">=0.86" so even if the provider workload is newer, it will be accepted by the consumer because ">=0.86".

The code you linked seems to confirm this:
```python
        spec = semver.SimpleSpec(self._normalized_version(needs))
        got = semver.Version.coerce(self._normalized_version(has))

        return spec.match(got)
```

My understanding of the provider version may be different (please correct me where I am wrong):
1. When the Provider is instantiated, the exact (and correct) version of the workload app needs to be passed to it.
2. Internally, the Provider lib can behave differently, depending on the version of the app.
3. Internally, the Consumer lib can behave differently, depending on the version of the app.
4. If the Consumer version string is conservative (i.e. not open ended such as ">=x.y.z") then the version should indeed be bumped.

rbarry82: 1. "needs" is very prescriptive. Again, the provider does not (and cannot) check the exact version of the workload app, since the OF does not contain a mechanism to know how to check this (API call? CLI flag? Buildinfo stored elsewhere? It's workload dependent). The version is _relatively_ arbitrary and can be used by you, the Charm author, to decide whether the provider and consumer are compatible. The guidance in the OF docs is not telling you "the operator framework will check this", it's telling you "you should [check this yourself](https://github.com/canonical/prometheus-operator/blob/c57508523dda7e6019ca718a133e6542b51fa44f/src/charm.py#L406) and [use it in the constructor](https://github.com/canonical/prometheus-operator/blob/c57508523dda7e6019ca718a133e6542b51fa44f/src/charm.py#L54) because someone may specify a different resource which makes your in-code constraints meaningless", not that the OF will check it.

2. It can behave differently, but should not. In practice, the reason for different `LIBAPI`, `LIBPATCH`, and `version` (in ProviderBase/ConsumerBase) is to specify a level of compatibility, where we can enforce that the provider/consumer have matching APIs and semantics, similar to `SemVer` in Golang and other implementations. If the Provider would need different behavior, constraints should be added so the consumer matches the expected behavior rather than having conditional behavior which depends on the version.

3. Consumer is essentially the same as 2.

4. Yes. It's worth pointing out, though, that the goal of using `SemVer` is to follow the Semantic Versioning paradigm. That is: the goal is to provide you, the library author, with flexibility in versioning to avoid "dependency hell" by providing some flexibility. The versioning scheme is completely up to you, as long as it is consistent. For whatever warts the system does have, the [go module versioning](https://blog.golang.org/publishing-go-modules) guidelines provide a good rationale under "semantic versioning and modules".

As "best practice", the charm code (not OF) should query the running application and use that in the constructor, not hardcoding it. My comment with comparators is saying "if you are going to hardcode it because AlertManager is prerelease (v0.x) software, the provider could be more flexible, and say `<=1.0` or similar". 

In the case of AlertManager, it's currently API v2. Whether you want to tie the provider/consumer to API version, config file format, explicit application version (by querying `/status` on the AlertManager endpoint), or something completely different is up to you.

lib/charms/karma_k8s/v0/karma.py
sed-i: docstings added here:
https://github.com/canonical/karma-operator/commit/fb49bd14172fd19e69a6a5170f4b6e42a47fc597

lib/charms/karma_k8s/v0/karma.py
sed-i: 1. Of course charm code is where you query for workload version. Here's an [example](https://github.com/canonical/alertmanager-operator/blob/main/src/charm.py#L110).

Now that we agree on (1) above, it seems even clearer than before that the argument to "version" is the actual version of the workload app.

afaiu, the interpretation "the provider can be a range" is inconsistent with the code snippet (and its intention) cited above (`SimpleSpec.match(Version.coerce(..))`).

rbarry82: Yes, it’s on the Karma library. 

We never disagreed on that point. My point is that semver provides flexibility alongside constraints. This PR neither queries nor uses the API versioning. 

A docstring does not overrule an industry-wide understanding of what semver is and what its uses are. Grafana, for example, is using the API version, not the workload version. 

The point of the comment is that a hard coded string to the version number does not enforce semver compliance in any meaningful way nor provide feedback about major/minor versioning changes to consumers. The last part of my last comment provided multiple options. 

Implementation is up to you, but it is not as prescriptive as asserted by any means.

rbarry82: A lot of this is defensive programming, and I mentioned the API version for a reason. Outside of the LMA use case, it does not cost a lot in code to say “what if?”

What if something uses the V1 API? How do we identify it? If the V2 API changes, how do we check?

If we don’t use API version, how do we know when the image updates or what is used to deploy if we don’t query?

sed-i: Using workload version is more defensive than API version, exactly for the "what ifs" you listed.

rbarry82: Just to close the loop (we discussed on Mattermost anyway). A dynamic version getter is great.

Using this version does not acknowledge the use case of something like `etcd`, where the “what if I want to use API `v2` no matter what the image is?” is covered. 

For “defensive” programming, “what if the image used changes?” and “what if the API used changes?” is a choice. Especially in the case of “generic” provider/consumer libs where it should fulfill, say, MariaDB || MySQL, which will have different image versions but more or less identical APIs for common use cases (similar also for galera, percona, pgpool, etc). One is not “more” defensive than the other. 

I’m not trying to be nitpicky here, but `SemVer` supports both use cases, and since this is a public discussion early in the history of `ProviderBase`, I want to be crystal clear that both options are essentially equally valid depending on the use case and intended purpose of the library, since these may get indexed. 

I approved this a long time ago anyway, and these comments are becoming circular. Let’s merge it.

PR #21, opened 2021-07-21
ops fixes

sed-i:
* use tox
* add charmcraft.yaml
* doc update about manual tests


charmcraft.yaml
```diff
@@ -0,0 +1,8 @@
+type: charm
```
mmanciop, line 1: I am not 100% sure this should be committed. @jnsgruk ?

metadata.yaml
```diff
@@ -1,5 +1,5 @@
 name: alertmanager-k8s
-display-name: Alertmanager
+display-name: Alertmanager K8S
```
mmanciop, line 2: Please no K8S in the display name, qualifying something for K8S is something we do only when necessary

mmanciop (changes requested):
I like the extra docu, comments on the rest

charmcraft.yaml
jnsgruk: Yeh this should definitely be committed; it will soon be mandatory in a later version of `charmcraft`. It's sparse at the moment, but will ultimately control the architectures the charm is built/run on.

metadata.yaml
jnsgruk: Agree.

sed-i: I was following an (semi?) established pattern: currently in charmhub there is "K8S" in alertmanager, prometheus, grafana, cassandra.

sed-i: Charmhub has the following alertmanager machine charms, so keep the "K8S" for this one?
https://charmhub.io/charmcraft-alertmanager
https://charmhub.io/prometheus-alertmanager

dstathis (approved):

src/charm.py
```diff
@@ -493,4 +493,4 @@ def _get_peer_addresses(self) -> List[str]:
 
 
 if __name__ == "__main__":
-    main(AlertmanagerCharm)
+    main(AlertmanagerCharm, use_juju_for_storage=True)
```
Abuelodelanada, line 496: Interesting, I never used `use_juju_for_storage`...

Side note here.
@jnsgruk we do not have this parameter documented in [either the SDK or Juju documentation](https://juju.is/docs/search?q=use_juju_for_storage). We only have a docstring in the [Operator framework code](https://operator-framework.readthedocs.io/en/latest/_modules/ops/main.html#main).

tox.ini
```diff
@@ -0,0 +1,47 @@
...
+select = E,W,F,C,N
+# Ignore W503, E501 because using black creates errors with this
+ignore = W503,E501
+
+[testenv:unit]
+deps = -r{toxinidir}/requirements.txt
+whitelist_externals = coverage
+commands =
+    coverage run --branch --source=src,lib/charms/alertmanager_k8s -m unittest -v
+    coverage report -m
```
Abuelodelanada, line 47: Nice to have tox!

CONTRIBUTING.md
```diff
@@ -65,6 +65,43 @@ Later on, upgrade packages as needed
 
     ./run_tests
 
+#### Manual testing
+Alerts can be created using alertmanager's HTTP API, 
+[for example](https://gist.github.com/cherti/61ec48deaaab7d288c9fcf17e700853a):
```
Abuelodelanada, line 70: I wouldn't link to a gist that can be deleted.

Abuelodelanada (approved):

src/charm.py
rbarry82: It's useful... basically always. `_should_use_controller_storage` should set the right default, but for K8s charms, it doesn't hurt anything to be explicit.

PR #22, opened 2021-07-28
add action show-config

sed-i:


mmanciop (approved):

actions.yaml
```diff
@@ -0,0 +1,4 @@
+show-config:
+  description: Show alertmanager config file.
+show-silences:
```
balbirthomas, line 3: English NitPick: "show-silences" or "show-silenced" ? Same about variable names below.

balbirthomas (approved):

actions.yaml
sed-i: this will show the "silences" not the silenced alarms. AM GUI has a tab named "silences" too

src/charm.py
```diff
@@ -46,18 +46,34 @@ def reload(self) -> bool:
...
 
-    def status(self) -> Optional[dict]:
-        url = urllib.parse.urljoin(self.base_url, "/api/v2/status")
+    @staticmethod
+    def _get(url: str) -> Optional[dict]:
         try:
             response = requests.get(url)
             if response.ok and response.status_code == 200:
-                status = json.loads(response.text)
+                text = json.loads(response.text)
```
Abuelodelanada, line 54: I think that checking both `response.ok` and `response.status_code == 200` [is redundant](https://github.com/psf/requests/blob/8149e9fe54c36951290f198e90d83c8a0498289c/requests/models.py#L696)

src/charm.py
```diff
@@ -46,18 +46,34 @@ def reload(self) -> bool:
...
+        return text
+
+    def status(self) -> Optional[dict]:
+        url = urllib.parse.urljoin(self.base_url, "/api/v2/status")
+        return self._get(url)
+
+    def silences(self, state: str = None) -> Optional[List[dict]]:
+        url = urllib.parse.urljoin(self.base_url, "/api/v2/silences")
+        silences = self._get(url)
+        logger.info("silences (all): %s", silences)
```
Abuelodelanada, line 69: Shouldn't be `logger.debug` ?

Abuelodelanada (approved):

src/charm.py
sed-i: removed "ok"

src/charm.py
sed-i: removed

PR #23, opened 2021-07-30
fix storage and charm upgrade

sed-i:
* add timeout for GET
* generate config for amtool
* update ip address on upgrade charm

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -448,6 +448,7 @@ class AlertmanagerCharm(CharmBase):
         # In the case of a single unit deployment, no 'RelationJoined' event is emitted, so
         # setting IP here.
         self._store_private_address()
+        self.provider.update_relation_data()
         self.karma_lib.target = self.api_address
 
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -49,7 +49,7 @@ def reload(self) -> bool:
     @staticmethod
     def _get(url: str) -> Optional[dict]:
         try:
-            response = requests.get(url)
+            response = requests.get(url, timeout=2.0)
```
Abuelodelanada, line 52: What about having this `timeout` value as a constant instead of hardcoded here?

Abuelodelanada (approved):

src/charm.py
sed-i: Good point! Fixed.

balbirthomas (approved):

balbirthomas (approved):

Abuelodelanada (approved):
LGTM

PR #24, opened 2021-08-10
fix unready bug and add unittests for consumer

### LINKED ISSUE #15: provider should NOT become unready on relation departed
https://github.com/canonical/alertmanager-operator/blob/489c32cc35131be42d601746ae169d6408cc1adb/src/charm.py#L431-L435

sed-i:
Fixes #15

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -43,7 +43,6 @@ ignore = W503,E501
 deps = 
   -r{toxinidir}/requirements.txt
   coverage
-# whitelist_externals = coverage
 commands =
     coverage run --branch --source=src,lib/charms/alertmanager_k8s -m unittest -v
     coverage report -m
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -40,8 +40,10 @@ select = E,W,F,C,N
 ignore = W503,E501
 
 [testenv:unit]
-deps = -r{toxinidir}/requirements.txt
-whitelist_externals = coverage
+deps = 
+  -r{toxinidir}/requirements.txt
```
balbirthomas, line 44: So you do not differentiate between requirements and dev requirements.

balbirthomas (approved):

tox.ini
sed-i: dev reqs are basically test reqs, and with tox it can all be captured under deps property. speaking of which, I should update the docs

Abuelodelanada (approved):
LGTM

PR #25, opened 2021-08-11
Patch Kubernetes service

jnsgruk:
Draft PR to patch the Kubernetes Service.

This is not finished, need to add unit tests and better exception handling for API calls, but this shows the basic method.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,6 +25,11 @@ logger = logging.getLogger(__name__)
 
 
 class K8sServicePatch:
+    namespace_file = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+    class PatchFailed(RuntimeError):
+        pass
+
     @staticmethod
     def namespace() -> str:
         """Read the Kubernetes namespace we're deployed in from the mounted service token
@@ -32,15 +37,15 @@ class K8sServicePatch:
         Returns:
             str: The current Kubernetes namespace
         """
-        with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r") as f:
+        with open(K8sServicePatch.namespace_file, "r") as f:
             return f.read().strip()
 
     @staticmethod
-    def _k8s_auth() -> bool:
+    def _k8s_auth():
         """Authenticate with the Kubernetes API using an in-cluster service token
 
-        Returns:
-            bool: True represents a successful authentication against the Kubernetes API
+        Raises:
+            K8sServicePatch.PatchFailed: if no permissions to read cluster role
         """
         # Authenticate against the Kubernetes API using a mounted ServiceAccount token
         kubernetes.config.load_incluster_config()
@@ -51,13 +56,11 @@ class K8sServicePatch:
             api.list_namespaced_service(namespace=K8sServicePatch.namespace())
         except kubernetes.client.exceptions.ApiException as e:
             if e.status == 403:
-                # If we can't read a cluster role, we don't have enough permissions
-                # Admin would need to run `juju trust` on this application to continue
-                # TODO inform caller that `juju trust` is needed
-                return False
+                raise K8sServicePatch.PatchFailed(
+                    "No permission to read cluster role. " "Run `juju trust` on this application."
+                ) from e
             else:
                 raise e
-        return True
 
     @staticmethod
     def _k8s_service(
@@ -93,19 +96,18 @@ class K8sServicePatch:
         )
 
     @staticmethod
-    def patch(app: str, service_ports: List[Tuple[str, int, int]]) -> bool:
+    def patch(app: str, service_ports: List[Tuple[str, int, int]]):
         """Patch the Kubernetes service created by Juju to map the correct port
 
         Args:
             app: app name
             service_ports: a list of tuples (name, port, target_port) for every service port.
 
-        Returns:
-            bool: True indicates successful patch
+        Raises:
+            K8sServicePatch.PatchFailed: if patching fails.
         """
         # First ensure we're authenticated with the Kubernetes API
-        if not K8sServicePatch._k8s_auth():
-            return False
+        K8sServicePatch._k8s_auth()
 
         ns = K8sServicePatch.namespace()
         # Set up a Kubernetes client
@@ -120,9 +122,7 @@ class K8sServicePatch:
                 namespace=ns, body=K8sServicePatch._k8s_service(app, service_ports)
             )
         except kubernetes.client.exceptions.ApiException as e:
-            logger.error("Failed to patch k8s service: %s", str(e))
-            return False
-        return True
+            raise K8sServicePatch.PatchFailed("Failed to patch k8s service: {}".format(e))
 
 
 class AlertmanagerAPIClient:
@@ -224,6 +224,7 @@ class AlertmanagerCharm(CharmBase):
 
         self._stored.set_default(
             pebble_ready=False,
+            k8s_service_patched=False,
             config_hash=None,
             launched_with_peers=False,
         )
@@ -537,6 +538,20 @@ class AlertmanagerCharm(CharmBase):
     def api_client(self) -> AlertmanagerAPIClient:
         return AlertmanagerAPIClient(self.private_address, self._api_port)
 
+    def _patch_k8s_service(self):
+        if self.unit.is_leader() and not self._stored.k8s_service_patched:
+            service_ports = [
+                (f"{self.app.name}-api", self.api_port, self.api_port),
+                (f"{self.app.name}-ha", self._ha_port, self._ha_port),
+            ]
+            try:
+                K8sServicePatch.patch(self.app.name, service_ports)
+            except K8sServicePatch.PatchFailed as e:
+                logger.error("Unable to patch the Kubernetes service: %s", str(e))
+            else:
+                self._stored.k8s_service_patched = True
+                logger.info("Successfully patched the Kubernetes service!")
+
     def _common_exit_hook(self) -> bool:
         if not self._stored.pebble_ready:
             self.unit.status = MaintenanceStatus("Waiting for pod startup to complete")
@@ -591,16 +606,7 @@ class AlertmanagerCharm(CharmBase):
 
     def _on_install(self, _):
         """Event handler for the install event during which we will update the K8s service"""
-        # At this point, it's debatable whether we should put the charm into a blocked state.
-        # My guess is right now, it's not essential, but as we start to wire these things more
-        # carefully, we should probably do so.
-        service_ports = [
-            (f"{self.app.name}-api", self.api_port, self.api_port),
-            (f"{self.app.name}-ha", self._ha_port, self._ha_port),
-        ]
-        if not K8sServicePatch.patch(self._container_name, service_ports):
-            logger.error("Unable to patch the Kubernetes service!")
-            self.unit.status = BlockedStatus("Run `juju trust` on this application to continue")
+        self._patch_k8s_service()
 
     def _on_peer_relation_joined(self, event: ops.charm.RelationJoinedEvent):
         self._common_exit_hook()
--- a/src/charm.py
+++ b/src/charm.py
@@ -560,7 +560,7 @@ class AlertmanagerCharm(CharmBase):
 
     def _patch_k8s_service(self) -> bool:
         """Patch the Kubernetes service created by Juju to map the correct port
-        
+
         Returns:
             bool: True indicates successful patch
         """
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -487,6 +488,14 @@ def _on_start(self, event: ops.charm.StartEvent):
         # available and the status was stuck on "Waiting for IP address". Adding this hook as a workaround.
         self._common_exit_hook()
 
+    def _on_install(self, _):
+        """Event handler for the install event during which we will update the K8s service"""
+        # At this point, it's debatable whether we should put the charm into a blocked state.
+        # My guess is right now, it's not essential, but as we start to wire these things more
+        # carefully, we should probably do so.
+        if not self._patch_k8s_service():
```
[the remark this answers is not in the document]
jnsgruk: Yeh so this could well end up us a small library or similar that could be shared across the O11y charms if it turns out to be a common enough requirement. This is just it's simplest implementation in many ways.

src/charm.py
```diff
@@ -487,6 +488,14 @@ def _on_start(self, event: ops.charm.StartEvent):
         # available and the status was stuck on "Waiting for IP address". Adding this hook as a workaround.
         self._common_exit_hook()
 
+    def _on_install(self, _):
+        """Event handler for the install event during which we will update the K8s service"""
+        # At this point, it's debatable whether we should put the charm into a blocked state.
```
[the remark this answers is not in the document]
jnsgruk: The "common exit hook" is really an anti-pattern I'd like to avoid (to be discussed in charm feedback sessions), precisely because of this sort of situation. I don't think this should be a case for `BlockedState`, but it should definitely log that something is up

src/charm.py
```diff
@@ -549,6 +558,94 @@ def _get_peer_addresses(self) -> List[str]:
...
+
+        try:
+            api.list_namespaced_service(namespace=self.namespace)
+        except kubernetes.client.exceptions.ApiException as e:
+            if e.status == 403:
+                # If we can't read a cluster role, we don't have enough permissions
+                self.unit.status = BlockedStatus(
+                    "Run `juju trust` on this application to continue"
+                )
+                return False
```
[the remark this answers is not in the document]
jnsgruk: Again, the common exit hook feels like a bad pattern, particularly for this reason

tests/test_charm.py
```diff
@@ -3,17 +3,19 @@
...
+from unittest.mock import Mock, patch
 
+import kubernetes
 import ops
+from charm import AlertmanagerAPIClient, AlertmanagerCharm
+from ops.model import BlockedStatus
 from ops.testing import Harness
 
-# from ops.model import ActiveStatus
+from .helpers import PushPullMock, patch_network_get, tautology
```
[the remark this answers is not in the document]
jnsgruk: `black -l99 src/` ;)

tests/test_charm.py
```diff
@@ -145,6 +153,76 @@ def test_show_silences_action_succeeding(self):
...
+    @patch("charm.AlertmanagerCharm._k8s_auth")
+    @patch("charm.kubernetes.client.CoreV1Api.delete_namespaced_service")
+    @patch("charm.kubernetes.client.CoreV1Api.create_namespaced_service")
+    def test_patch_k8s_service(self, create, delete, auth, ns):
+        ns.return_value = "lma"
+        name = self.harness.charm.app.name
+        create.return_value = delete.return_value = auth.return_value = True
+        patched = self.harness.charm._patch_k8s_service()
+        delete.assert_called_with(name=name, namespace=self.harness.charm.namespace)
+        create.assert_called_with(
```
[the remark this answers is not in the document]
jnsgruk: I think so; as part of the tests we have:

1. ensured that the `k8s_service` property returns a sane definition of a service
2. ensured that the logic to create the service uses that sane definition

tests/test_charm.py
```diff
@@ -145,6 +153,76 @@ def test_show_silences_action_succeeding(self):
...
+        authed = self.harness.charm._k8s_auth()
+        list_svc.assert_called_with(namespace=self.harness.charm.namespace)
+        self.assertEqual(authed, True)
+
+        # Now test what happens when listing a svc throws an exception
+        list_svc.side_effect = kubernetes.client.exceptions.ApiException(status=403)
+        authed = self.harness.charm._k8s_auth()
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus("Run `juju trust` on this application to continue"),
```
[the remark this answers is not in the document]
jnsgruk: Yeh, I'm easy either way on that

jnsgruk:
@sed-i My intention was for you to take this on going forward, you should have the ability to push to the branch as a collaborator. I wanted to give you just enough to get started and at least sketch some of the tests out for you!

src/charm.py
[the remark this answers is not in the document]
sed-i: Refactored. WDYT?

src/charm.py
```diff
@@ -1,29 +1,125 @@
...
+            # Set up a Kubernetes client
+            api = kubernetes.client.CoreV1Api(kubernetes.client.ApiClient())
+            # Delete the existing service so we can redefine with correct ports
+            # I don't think you can issue a patch that *replaces* the existing ports,
+            # only append
+            api.delete_namespaced_service(name=app, namespace=ns)
+            # Recreate the service with the correct ports for the application
+            api.create_namespaced_service(
+                namespace=ns, body=K8sServicePatch._k8s_service(app, service_ports)
+            )
```
jnsgruk, line 118: One thing I didn't get around to doing was wrapping this with a try/except and catch some likely exceptions. I can't see a normal situation where this is likely to fail, but nonetheless worth catching so it doesn't just blindly raise. If you see in the auth method, the Kubernetes `APIException` might be the one to catch, and it carries with it some other attributes (like `status` etc.)

sed-i: Both here and in _auth, the immediately obvious exceptions are `ApiTypeError` and `ApiValueError` (neither of which inherit `ApiException`. What could I do under `except` other than `logger.error`?

BTW, all the exception attributes (e.g. `status`, `reason`) are already incorporated into its `__str__`:
```python
    def __str__(self):
        """Custom error messages for exception"""
        error_message = "({0})\n"\
                        "Reason: {1}\n".format(self.status, self.reason)
```

tests/test_charm.py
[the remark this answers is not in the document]
sed-i: black does _not_ reorder imports for me

src/charm.py
```diff
@@ -1,29 +1,130 @@
...
 
+class K8sServicePatch:
+    @staticmethod
+    def namespace() -> str:
+        """Read the Kubernetes namespace we're deployed in from the mounted service token
+
+        Returns:
+            str: The current Kubernetes namespace
+        """
+        with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r") as f:
```
Abuelodelanada, line 35: I would put the path in a constant to make it easier to modify it in the future

src/charm.py
```diff
@@ -1,29 +1,130 @@
...
+        api = kubernetes.client.CoreV1Api(kubernetes.client.ApiClient())
+
+        try:
+            api.list_namespaced_service(namespace=K8sServicePatch.namespace())
+        except kubernetes.client.exceptions.ApiException as e:
+            if e.status == 403:
+                # If we can't read a cluster role, we don't have enough permissions
+                # Admin would need to run `juju trust` on this application to continue
+                # TODO inform caller that `juju trust` is needed
+                return False
```
Abuelodelanada, line 57: What about logging this with a `.warning()`?

Abuelodelanada (approved):

src/charm.py
sed-i: Done.

src/charm.py
sed-i: Replaced retvals with exceptions.
Trying to avoid using the `logging` module in "generic" classes.
Please re-review.

src/charm.py
```diff
@@ -1,29 +1,130 @@
...
-from typing import Dict, List, Optional, Any
+import utils
 
 logger = logging.getLogger(__name__)
 
 
+class K8sServicePatch:
+    namespace_file = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+    class PatchFailed(RuntimeError):
```
balbirthomas, line 30: I don't see why you would like an exception type to be a nested class.

src/charm.py
```diff
@@ -1,29 +1,130 @@
...
-import yaml
-import json
-import logging
-from typing import Dict, List, Optional, Any
+import utils
 
 logger = logging.getLogger(__name__)
 
 
+class K8sServicePatch:
```
balbirthomas, line 27: It will be useful to have some docstrings on what is the responsibility of this class, and how it should be used.

src/charm.py
```diff
@@ -1,29 +1,130 @@
 #!/usr/bin/env python3
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
+import json
```
balbirthomas, line 4: Do you really want all this code in you charm.py ? Alternatively does it make sense to factor it out into a module (say kubernetes_service.py) of its own.

src/charm.py
```diff
@@ -1,29 +1,130 @@
...
+                labels={"app.kubernetes.io/name": app},
+            ),
+            spec=kubernetes.client.V1ServiceSpec(
+                ports=ports,
+                selector={"app.kubernetes.io/name": app},
+            ),
+        )
+
+    @staticmethod
+    def patch(app: str, service_ports: List[Tuple[str, int, int]]):
```
balbirthomas, line 99: If this function is meant only to patch a port, wouldn't it be better to call it something like `patch_port()`.

src/charm.py
```diff
@@ -435,6 +538,20 @@ def api_address(self):
     def api_client(self) -> AlertmanagerAPIClient:
```
balbirthomas, line 538: docstring would be nice

src/charm.py
```diff
@@ -435,6 +538,20 @@ def api_address(self):
...
         return AlertmanagerAPIClient(self.private_address, self._api_port)
 
+    def _patch_k8s_service(self):
```
balbirthomas, line 541: docstring would be nice

balbirthomas (approved):

src/charm.py
jnsgruk: This isn't a bad idea, given that other folk in the O11y team will need this too. I don't think for now this should be a charm library because we will ultimately fix this in Juju, but worth pulling out into its own file with it's own tests probably.

src/charm.py
jnsgruk: +1, consider docstring style too (Google style)

src/charm.py
jnsgruk: Perhaps `set_ports`?

Also as discussed yesterday, perhaps a class called `PortMapping` with properties for `from_port`, `to_port`, `name` with the `__repr__` and `__dict__` overriden here could result in tider type hints?

tests/test_charm.py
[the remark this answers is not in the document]
jnsgruk: Oh hang on, that'll be `Pylance` which is the new default language server for Python in VSCode.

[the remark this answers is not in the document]
jnsgruk: Now this has come up, perhaps we should standardise in the team to stop this sort of stuff happening. We already have `black` and `flake8` as part of the `tox` setup, perhaps adding `isort` too isn't a bad idea? Ideal flow I guess would be

`flake8` -> `isort` -> `black`?

Abuelodelanada (commented):
Looks good to me, but I like Balbirs' suggestions

src/charm.py
sed-i: `class K8sServicePatch` is essentially just a namespace for related stuff, that's why the nesting.
The "PatchFailed" exception makes no sense anywhere else outside.
Now that I'll be moving it to its own file/module (i.e. namespace) I will take it out.

tests/test_charm.py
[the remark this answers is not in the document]
sed-i: Sounds good. I will look into adding `isort --diff` to `tox -e lint` and `isort src/. tests/. lib/.../.` to `tox -e prettify` in a separate PR.

src/charm.py
sed-i: Done

src/charm.py
sed-i: Done

src/charm.py
sed-i: Done

src/charm.py
sed-i: Done

src/charm.py
sed-i: Done

Abuelodelanada (approved):

PR #27, opened 2021-08-19
List also a UI port in the K8s service

mmanciop:
While it is true that API and UI port in Alertmanager are the same, this is not self-evident and a seldom known to users (who almost never interact with the alertmanager API, but rather the UI).

This commit lists a dedicated `-ui` port for Alertmanager, with the same `9093` port as the `-api` one.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -443,6 +443,7 @@ class AlertmanagerCharm(CharmBase):
         """Fix the Kubernetes service that was setup by Juju with correct port numbers"""
         if self.unit.is_leader() and not self._stored.k8s_service_patched:
             service_ports = [
+                (f"{self.app.name}-ui", self.api_port, self.api_port),
                 (f"{self.app.name}-api", self.api_port, self.api_port),
                 (f"{self.app.name}-ha", self._ha_port, self._ha_port),
             ]


src/charm.py
```diff
@@ -443,6 +443,7 @@ def _patch_k8s_service(self):
         """Fix the Kubernetes service that was setup by Juju with correct port numbers"""
         if self.unit.is_leader() and not self._stored.k8s_service_patched:
             service_ports = [
+                (f"{self.app.name}-ui", self.api_port, self.api_port),
                 (f"{self.app.name}-api", self.api_port, self.api_port),
```
jnsgruk, line 447: Hmm, I think I'd prefer this:

```suggestion
                (f"{self.app.name}", self.api_port, self.api_port),
```

jnsgruk (changes requested):
For discussion.

Also @sed-i I was going to comment and say this will likely break tests, but it doesn't... and it probably should! Can be addressed in a seperate PR.

mmanciop:
Unfortunately this cannot work, we can have only one port with the same value :-)

sed-i:
> this will likely break tests, but it doesn't... and it probably should!

@jnsgruk - which part? all I see in the diff is addition of `f"{self.app.name}-ui"`.
Did you mean a test to check that there are in fact only two ports listed and that their names and ports are such and such?

jnsgruk:
Yeh, I'd assume that the tests were trying to validate that set port/name was created as part of a hook?

sed-i:
Wouldn't it just be an exact snapshot of the code itself to compare against? Basically a coded formalization of `# don't change this` comment. Is that the intention?

PR #28, opened 2021-08-19
Improve K8s service patching

mmanciop:
* Rename the `alertmanager-api` port to `alertmanager` for clarity, as the same `9093` port serves both UI and API
* Ensure K8s service is patched on charm upgrade

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -123,7 +123,6 @@ class AlertmanagerCharm(CharmBase):
 
         self._stored.set_default(
             pebble_ready=False,
-            k8s_service_patched=False,
             config_hash=None,
             launched_with_peers=False,
         )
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):

sed-i (changes requested):
As part of this change, can you please get rid of `_stored.k8s_service_patched`?
It seems that we only call it when it needs to be called anyway.
I originally had `k8s_service_patched` when I was calling the patch from within _common_exit_hook.

jnsgruk (approved):
Perhaps one day we could add some code to the patcher class that says "check if I'm patched" by just fetching the V1Service from the K8s api and comparing with one that we construct. That said, the current operation is fast, and would likely only result in a *very* short service disruption (likely in the ms range) - and only if someone issues an upgrade, which will restart alertmanager anyway...

sed-i (approved):

PR #29, opened 2021-08-23
add pushover, webhook receivers, using namespaced config

sed-i:


src/config.py
```diff
@@ -0,0 +1,71 @@
+import abc
+from typing import Optional
+
+
+class ConfigBase(abc.ABC):
```
balbirthomas, line 5: docstrings ? : What is this class for ? How is it intended to be used ?

src/config.py
```diff
@@ -0,0 +1,71 @@
...
+    def from_dict(data: dict) -> Optional[dict]:
+        ...
+
+    @staticmethod
+    @abc.abstractmethod
+    def is_valid(data: dict) -> bool:
+        ...
+
+
+class ReceiverBase(ConfigBase):
```
balbirthomas, line 18: docstring ?

balbirthomas (approved):
Approving. But docstrings would be very useful for each of your classes explaining what they are meant for and how they are intended to be used.

src/config.py
sed-i: Added

src/config.py
sed-i: Added

Abuelodelanada (approved):
LGTM

PR #30, opened 2021-08-24
jnsgrukification

sed-i:


src/alertmanager_client.py
```diff
@@ -0,0 +1,74 @@
...
+import logging
+import urllib.parse
+from typing import List, Optional
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class AlertmanagerAPIClient:
```
jnsgruk, line 15:
```suggestion
class Alertmanager:
```

src/alertmanager_client.py
```diff
@@ -0,0 +1,74 @@
...
+    def reload(self) -> bool:
+        """Send a POST request to to hot-reload the config.
+        This reduces down-time compared to restarting the service.
+
+        Returns:
+          True if reload succeeded (returned 200 OK); False otherwise.
+        """
+        url = urllib.parse.urljoin(self.base_url, "/-/reload")
+        try:
+            response = requests.post(url, timeout=self.timeout)
```
jnsgruk, line 31:
```suggestion
            response = urllib.request.urlopen(url, data=None, timeout=self.timeout)
```

Perhaps drop the dependency on `requests`?

sed-i: Done

src/alertmanager_client.py
sed-i: Done

jnsgruk (approved):
LGTM! :)

PR #32, opened 2021-09-02
invert karma relation

sed-i:


lib/charms/karma_k8s/v0/karma.py
```diff
@@ -136,50 +134,49 @@ class KarmaProvider(ProviderBase):
...
     ```python
-    from charms.alertmanager_karma.v0.karma import KarmaProvider
+    from charms.alertmanager_karma.v0.karma import KarmaConsumer
     ```
 
     In your charm's `__init__` method:
 
     ```python
-    self.provider = KarmaProvider(
+    self.karma_lib = KarmaConsumer(
```
balbirthomas, line 143: Nitpick: odd to name an object as a library.

balbirthomas (approved):

PR #34, opened 2021-09-08
Add integration tests

sed-i:


sed-i:
To run lma bundle tests in other charms, one would clone duplicate the `test_lma_bundle.py` file and update `tox.ini` as in this PR.

See CONTRIBUTING in the accompanying PR for the full picture:
https://github.com/canonical/lma-light-bundle/pull/4

balbirthomas (approved):

metadata.yaml
```diff
@@ -27,9 +27,8 @@ containers:
 resources:
   alertmanager-image:
     type: oci-image
-    # The officially supported image is ubuntu/prometheus-alertmanager
     description: OCI image for alertmanager
-
+    upstream-source: ubuntu/prometheus-alertmanager
```
jnsgruk, line 31: This is a deprecated field aiui

knkski: It's not a deprecated field AFAIK. We use it extensively in the Kubeflow bundle. Juju itself won't do anything with the field, but it a) acts as a comment about where someone should look if they want to inspect the Dockerfile for the charm they're running, and b) is also used in a lot of tooling around charm deployment and publishing.

See here for how we use that field when publishing a charm: https://github.com/canonical/kubeflow-dashboard-operator/blob/798e432c/.github/workflows/publish.yaml#L33

When Launchpad starts building charms for charmhub.io, we'll want some sort of similar functionality, and so may want to just use this existing field.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test):
+    """Build the charm-under-test and deploy it together with related charms.
+
+    Assert on the unit status before any relations/configurations take place.
+    """
+    # build and deploy charm from local source folder
+    charm_under_test = await ops_test.build_charm(".")
+    resources = {"alertmanager-image": "ubuntu/prometheus-alertmanager"}
+    await ops_test.model.deploy(charm_under_test, resources=resources, application_name="am")
```
knkski, line 29: This is where `upstream-source` is really useful, as you don't have to specify the image being used again. I'm also bugging @johnsca to add some functionality to `pytest-operator` where you could write something like this:

```python
charm = await ops_test.build_charm()
await ops_test.model.deply(charm, resources=charm.default_resources())
```

metadata.yaml
knkski: Also, this field should probably have a version specifier such as `0.21-20.04_beta`, or at least a sha hash if we want to use the current latest image

metadata.yaml
```diff
@@ -27,9 +27,8 @@ containers:
...
 provides:
   alerting:
     # The provider (alertmanager) adds the following key-value pair to the relation data bag of
```
knkski, line 34: You may want to check out this library, that makes adding a schema to relations like this very simple:

https://github.com/canonical/serialized-data-interface

Then we don't have to have comments like this, as the schemas are quite readable on their own. As an example, here's the schema for the `ingress` relation used in Kubeflow:

https://github.com/canonical/operator-schemas/blob/master/ingress.yaml

metadata.yaml
jnsgruk: Yeh, I guess for charms published to old charm store it makes sense, though I suspect the O11y team will never publish there. Interesting point about encapsulating the information in the field to make the automation tooling easier for publishing though, that is a nice point.

Even though the workflow is to upload an oci-resource, then associate that with a release, I could see why having this in metadata.yaml would make it easier to automate.

tests/integration/helpers.py
```diff
@@ -0,0 +1,50 @@
...
+        run_args.append(f"--channel={channel}")
+
+    retcode, stdout, stderr = await ops_test.run(*run_args)
+    assert retcode == 0, f"Deploy failed: {(stderr or stdout).strip()}"
+    log.info(stdout)
+    await ops_test.model.wait_for_idle(apps=[alias], status=wait_for_status, timeout=60)
+
+
+async def get_unit_address(ops_test, app_name: str, unit_num: int) -> str:
+    # return ops_test.model.applications[app_name].units[unit_num].data["private-address"]
```
knkski, line 34: Should this be removed?

tests/integration/test_charm.py
knkski: Also, since you're already parsing the `metadata.yaml` field as `METADATA`, you could just look up `upstream-source` from that instead of respecifying it here, as that tends to get forgotten about when updating `metadata.yaml`

metadata.yaml
sed-i: Yes, `upstream-source` is the only consistent machine-readable place to document the image, which is also useful for local automation.

metadata.yaml
sed-i: I really like the schema idea.
Going all-in with SDI would need a PR of its own.
I also heard that relation schemas are coming soon from juju itself..?

knkski: I'm not sure that what is on the roadmap matches what SDI provides, though @jnsgruk can speak more to that

jnsgruk: My understanding is this:

- We have some underlying work to do in Juju core to enable some sort of execution environment, which would indeed enable validation/enforcement of schemas (this is the Lua stuff you've heard about)
- Once that is in place, we will develop it further to enable schema validation on config, relation databags, etc. (implied here is more complex, nested config/databags)

I think in light of this, where SDI is already in use, why not?! It works, it solves a problem for you guys. For O11y, it may just be prudent to wait out, but I'm certainly open to the discussion.

tests/integration/test_charm.py
sed-i: Thanks! Done.

tests/integration/helpers.py
sed-i: Removed.

metadata.yaml
sed-i: > this field should probably have a version specifier such as `0.21-20.04_beta`

Good point! We haven't been freezing versions at all so far. One difficulty is remembering to match it with the (opaque) uploaded image resource on charmhub.

tests/integration_lma/test_lma_bundle.py
```diff
@@ -0,0 +1,53 @@
...
+                branch="feature/integration_tests",  # TODO change to main after merged
+                depth=1,
+            )
+
+            run_args = [
+                "pytest",
+                "-v",
+                "--tb",
+                "native",
+                "--log-cli-level=INFO",
```
knkski, line 45: This should get picked up already from the options defined in `pyproject.toml`, shouldn't it?

tox.ini
```diff
@@ -77,3 +77,24 @@ commands =
...
+[testenv:integration]
+description = Run integration tests
+deps =
+    git+https://github.com/juju/python-libjuju.git
+    pytest
+    git+https://github.com/charmed-kubernetes/pytest-operator.git
+commands =
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {toxinidir}/tests/integration
+
+[testenv:integration-lma]
```
knkski, line 90: I don't know how big the lma-light or lma-ha bundles are going to get, but you may eventually not be able to run the full integration tests in this repo. The Charmed Kubeflow bundle is big enough that it costs us money to run the full and lite test suites because we have to spin up an external VM, so we only run it in the Charmed Kubeflow CI. One of the sub-bundles in it is Kubeflow Pipelines, which by itself almost hits the Github Actions runner resource limits, and doesn't actually have all that many charms in it:

https://github.com/canonical/kfp-operators/blob/master/bundle.yaml

If you did pull bundle-level integration testing out of this repo, one thing you could do to mitigate the impact is have CI for the bundle run on a schedule, like we do for testing Charmed Kubeflow on AKS:

https://github.com/canonical/bundle-kubeflow/blob/master/.github/workflows/test-aks.yaml

Which you may want to do anyways, as that will catch any dependencies like Juju or MicroK8s changing unexpectedly, which is an issue we've had for the Charmed Kubeflow bundle. Since most of the PRs and general activity happen in repos for sub-bundles like https://github.com/canonical/kfp-operators/, the CI for https://github.com/canonical/bundle-kubeflow/ is run irregularly, and we've been meaning to put in place more scheduled CI there as well.

metadata.yaml
```diff
@@ -27,9 +27,8 @@ containers:
 resources:
```
knkski, line 27: This PR doesn't display this part of the file, so I'll just comment here, but it looks like `display-name` isn't supported in Metadata V2 charms / by charmhub.io:

https://discourse.charmhub.io/t/3674/24

Juju just ignores extra fields, so it won't break anything, but I don't think it will do anything anymore.

sed-i: Thanks for pointing this out. Removed.

tox.ini
sed-i: The intention here is to be able to run integration tests both locally and as part of CI.

> If you did pull bundle-level integration testing out of this repo, one thing you could do to mitigate the impact is have CI for the bundle run on a schedule

Good point. Although if the lma-bundle repo is the one that would be running tests on a schedule, it would still need to checkout the main branch of this repo, build the charm etc., so I'm not sure if it is advantageous. I think that running tests with obvious/correct context has value/increases merge responsibility.
Something in juju/microk8s etc changing is a good use case though. I will add it to the bundle repo shortly.
Some of such surprises could be mitigated by version freezing of the pypi and oci images.

tests/integration/helpers.py
```diff
@@ -0,0 +1,49 @@
...
+    log.info(stdout)
+    await ops_test.model.wait_for_idle(apps=[alias], status=wait_for_status, timeout=60)
+
+
+async def get_unit_address(ops_test, app_name: str, unit_num: int) -> str:
+    status = await ops_test.model.get_status()  # noqa: F821
+    return status["applications"][app_name]["units"][f"{app_name}/{unit_num}"]["address"]
+
+
+async def update_status_freq(ops_test, interval: str = "10s"):
```
dstathis, line 38: This can be done with `await ops_test.model.set_config({"update-status-hook-interval": "10s"})`

tests/integration/helpers.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from pathlib import Path
+
+log = logging.getLogger(__name__)
+
+
+async def cli_deploy_and_wait(
```
dstathis, line 10: Is this function used anywhere?

metadata.yaml
dstathis: What is the reason for juju not reading this? It would be really nice to not have to specify --resource every time I deploy a local charm.

tests/integration_lma/test_lma_bundle.py
```diff
@@ -0,0 +1,53 @@
...
+        # https://github.com/charmed-kubernetes/pytest-operator/blob/main/pytest_operator/plugin.py
+        # Copying the built charm elsewhere to make sure it is still available when the lma bundle
+        # is being deployed.
+        shutil.copy(charm_under_test, temp_charm_dir)
+        charm_under_test = os.path.join(temp_charm_dir, os.path.split(charm_under_test)[-1])
+
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as temp_dir:
+            log.info("Cloning lma-light-bundle repo to a temp folder %s", temp_dir)
+            Repo.clone_from(
+                "https://github.com/canonical/lma-light-bundle.git",
```
dstathis, line 33: This needs to be parameterized. If I want to run this offline I need to be able to at least point to a local copy of the repo.

tests/integration_lma/test_lma_bundle.py
```diff
@@ -0,0 +1,53 @@
...
+import logging
+import os
+import shutil
+import tempfile
+
+import pytest
+from git import Repo
+
+log = logging.getLogger(__name__)
+
```
dstathis, line 14: This is a comment on the whole file. If I understand correctly, this appears to be a pytest test which clones the repo we want and runs pytest against it. Why are we doing this in a test. Is it not possible to add the following to tox.ini?
```
git clone ...
pytest ...
```
Besides most likely leading to bugs, this is just incredibly confusing for no apparent benefit. If there is a reason here that I don't know about, please at least put some comments at the top of the file explaining the need for it.

dstathis (changes requested):

tests/integration_lma/test_lma_bundle.py
dstathis: Upon further reflection, `charmcraft build` also requires network. I would like to have offline tests working, but it won't be a blocker for me as it is a significant amount of work to do so.

tests/integration/helpers.py
sed-i: not anymore! thanks

tests/integration/helpers.py
sed-i: thanks!

metadata.yaml
sed-i: @dstathis for now you'd have to use yq, for example: `yq eval '.resources[].upstream-source' metadata.yaml`, assuming you only have one image.

tests/integration_lma/test_lma_bundle.py
sed-i: I had tox taking most of the burden before switching to this pattern.
Apart from git clone and pytest, there is also `charm_under_test = await ops_test.build_charm(".")`, and to replace that I would have to call charmcraft from tox. [It was ugly](https://github.com/canonical/alertmanager-operator/pull/34/commits/feb05f9d614fcb9defd2ff46cbf0a9f483a49cb0).

knkski:
@dstathis not sure why I can't respond to your comment inline, so responding here.

> What is the reason for juju not reading this? It would be really nice to not have to specify --resource every time I deploy a local charm.

💯  on that. `upstream-source` was accepted as a compromise a while back as a field that Juju ignores, but a lot of the tooling for Charmed Kubeflow operates with. @jnsgruk and @jameinel can speak to why that compromise was made, and/or if it's something we can revisit.

tests/integration_lma/test_lma_bundle.py
dstathis: Two points here:
1. I don't think that was ugly.
2. Why can't you still use ops_test.build_charm in the integration tests? You are calling pytest still.

sed-i: **1**
I think you overlooked [this line](https://github.com/canonical/alertmanager-operator/blob/feb05f9d614fcb9defd2ff46cbf0a9f483a49cb0/tox.ini#L103):
`tmpdir = /tmp/pseudo_random/scf3u8/in_lieu_of_mktemp_d`

Also for that to work I needed to roll my own `render_bundle.py`.
Also, for cleanup, there is an `rm -rf` in there, which is ugly on its own.

**2**
Indeed, now I build the charm inside the python code. But that also means that I need to call pytest from within the python code, because tox wouldn't know the path of the *.charm (unless I add even more ugliness).
Basically it's either split the test logic into tox + py + ugliness, or keep it all inside the py. I think the keeping it all inside the py makes the most sense.

dstathis: Sorry to be a thorn in your side here but if this is going to end up in every repo I feel I must.

1. I don't think using a tmpdir is an issue. Are you concerned that you don't get the same safety as the python tempfile library? If so, the command `mktemp` is almost certainly in the github containers and does the same thing.

Haven't you made a render_bundle.py already?

`rm` is how you delete files on linux. I'm not sure what the issue is with running it

2. I'm not really following here. What do you mean tox wouldn't know the path? If pytest builds the charm and runs tests on it, why would tox need to know anything about it?

mmanciop: In my experience, tmpdirs are far better at avoiding dirty-state bugs, and markedly worse for debuggability. Which trade off  to take here depends on the likelihood and possible severity of dirty-state bugs.

dstathis: 1
Good point about storing the variable. I probably would make a shell script or python script.

>> Besides most likely leading to bugs, this is just incredibly confusing for no apparent benefit.
> This argument is vague so please defend it, otherwise I will resolve :)

It is very confusing the way it currently works. We are calling a pytest test sweet that has a single instance of a test. That test deploys a bundle then initiates another call of pytest which itself has a bunch of tests. What is the significance of the results of the outer test vs the inner tests? Why do we have `assert 5 == 7`? Does this mean the outer test always fails?

When looking at this design there are just so many questions and from the outside looking in it looks like we are using pytest to run code rather than run tests.

2
I think we are in agreement on this thread. Tox does not need to build the charms/bundle.

sed-i: > Why do we have assert 5 == 7?

Oops! It was for sanity check.

> it looks like we are using pytest to run code rather than run tests.

It's a bit more subtle: the "outer pytest" also builds the charm locally which is worthy of being under pytest, and, yes, does some file ops.
This goes back to the tradeoff between:
- Everything is inside pytest py code, and enjoys the built-in ops_test context, and works exactly the same locally and on ghwf; vs
- Use a shell script (or [equivalent](https://github.com/canonical/alertmanager-operator/blob/feb05f9d614fcb9defd2ff46cbf0a9f483a49cb0/tox.ini) of) to git clone, sg lxd charmcraft pack etc., splitting test logic across tox -> shell script.

PR #35, opened 2021-09-10
remove ops.relation refs

sed-i:


lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -43,7 +42,25 @@ def restore(self, snapshot):
...
 
 
-class AlertmanagerConsumer(ConsumerBase):
+class AlertmanagerConsumerEvents(ObjectEvents):
+    """Event descriptor for events raised by `AlertmanagerConsumer`."""
+
+    cluster_changed = EventSource(ClusterChanged)
+
+
+class RelationManagerBase(Object):
```
balbirthomas, line 51: It is not clear to me why you need this class.

lib/charms/karma_k8s/v0/karma.py
```diff
@@ -165,22 +174,20 @@ class KarmaConsumer(ConsumerBase):
...
-    def __init__(self, charm, name: str, consumes: dict, multi: bool = False):
-        super().__init__(charm, name, consumes, multi)
+    def __init__(self, charm, name: str):
+        super().__init__(charm, name)
         self.charm = charm
 
         events = self.charm.on[self.name]
-        self.framework.observe(events.relation_changed, self._on_relation_changed)
-        self.framework.observe(events.relation_departed, self._on_relation_departed)
+        self.charm.framework.observe(events.relation_changed, self._on_relation_changed)
```
balbirthomas, line 189: `Object` also has a `framework` instance. I am not entirely sure if this change is necessary.

balbirthomas (approved):

lib/charms/karma_k8s/v0/karma.py
sed-i: Leftover. Fixed.

lib/charms/alertmanager_k8s/v0/alertmanager.py
sed-i: Same is repeated for both provider and consumer. Reduce duplication and (re)establish contract.

PR #39, opened 2021-10-05
Upkeep

sed-i:
This PR contains several upkeep changes that I encountered in other charms and projects:
- [x] add spelling check (codespell) to the lint tox env
- [x] add CodeQL workflow
- [x] better handling of relation name: add default and validation

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,64 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [main]
+
+permissions:
+  security-events:
+    write
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['python']
+        # Learn more...
+        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file. 
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
+
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1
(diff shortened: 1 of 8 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,64 @@
...
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [main]
```
mmanciop, line 13: I think it might make sense to run this check on any PR, irrespective of the target branch (these are github workflows we do not current use, but still)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,64 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
```
mmanciop, line 8: I wonder if we should run it also nightly on main. You know, just in case one does a `git push` in the wrong direction :P

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -70,10 +69,21 @@ class RelationManagerBase(Object):
         name (str): consumer's relation name
     """
 
-    def __init__(self, charm: CharmBase, relation_name):
+    def __init__(self, charm: CharmBase, relation_name: str, relation_role: RelationRole):
```
mmanciop, line 72: Not checking the interface? -1

mmanciop (changes requested):
THe validation of the relation should include the interface name.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -70,10 +69,21 @@ class RelationManagerBase(Object):
...
         self.name = relation_name
 
+    def _validate_relation(self, relation_name: str, relation_role: RelationRole):
+        try:
+            if self.charm.meta.relations[relation_name].role != relation_role:
+                raise ValueError(
+                    f"Invalid role for relation: {relation_name}; must be '{relation_role}'"
+                )
+        except KeyError:
+            raise ValueError(f"Invalid relation name: {relation_name}")
```
mmanciop, line 85: This is not informative as an error message: it should read something like "Relation XYZ not found in the charm's metadata.yaml"

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -70,10 +69,21 @@ class RelationManagerBase(Object):
...
         super().__init__(charm, relation_name)
+        self.charm = charm
+        self._validate_relation(relation_name, relation_role)
         self.name = relation_name
 
+    def _validate_relation(self, relation_name: str, relation_role: RelationRole):
+        try:
+            if self.charm.meta.relations[relation_name].role != relation_role:
+                raise ValueError(
+                    f"Invalid role for relation: {relation_name}; must be '{relation_role}'"
```
mmanciop, line 82: This is not informative as an error message: it should read something like "Relation XYZ in the charm's metadata.yaml should be {provided|required} to be managed by this library, but instead it is {required|provided}"

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
sed-i: My impression was that hardcoding the interface here is still debatable :)

mmanciop: Your impression is likely wrong ;-)
Yesterday I showed how you cannot rely on the interface in `metadata.yaml`, remember?

.github/workflows/codeql-analysis.yml
sed-i: Not sure. Let's wait for additional input?

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
sed-i: Yes. I am not convinced yet.
Although if we already hardcode the relation name there's not much sense to keep the interface name open.

mmanciop: I think you are mixing things up a bunch: the interface name is fixed because otherwise `juju relate` simply does not work. The relation name is something that, as discussed in the last plenary, makes sense fixed for things like observability relations, and not for things like database relations.

sed-i: Added.

Abuelodelanada (approved):

pyproject.toml
```diff
@@ -1,12 +1,5 @@
-[project]
-name = "alertmanager-operator"
-description = "Kubernetes charm for Alertmanager"
-license = "Apache 2.0"
-python = "^3.8"
-readme = "README.md"
-homepage = "https://charmhub.io/alertmanager-k8s"
-repository = "https://github.com/canonical/alertmanager-operator"
-documentation = "https://charmhub.io/alertmanager-k8s/docs"
```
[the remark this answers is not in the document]
sed-i: No, these entries are optional.
I would have kept them if they weren't duplicating what we already have in metadata.yaml and tox.ini.

.github/workflows/codeql-analysis.yml
sed-i: #42

mmanciop (approved):

PR #41, opened 2021-10-06
User provides config

sed-i:
This PR fundamentally changes the way the alertmanager charm is configured:
- [x] remove all existing config options
- [x] add two config items: `config_file` and `templates_file`

This way the user (admin) has full control (max flexibility) over the config.


config.yaml
```diff
@@ -2,34 +2,19 @@
...
+      Alertmanager configuration file (yaml), with the exclusion of the templates section.
+      Refer to https://www.prometheus.io/docs/alerting/latest/configuration/ for full details.
     default: ""
-  # pushover config
-  pushover::send_resolved:
-    type: boolean
-    description: Whether or not to notify about resolved alerts (default = true).
-    default: true
-  pushover::user_key:
+  templates_file:
```
mmanciop, line 11: I wonder if we should call it `templates` and `config` instead. Ofc if we do it like currently in the PR, when Juju finally has structured configs, we still have the cool config names available. I am conflicted.

src/alertmanager_client.py
```diff
@@ -33,21 +34,42 @@ def reload(self) -> bool:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = urllib.parse.urljoin(self.base_url, "/-/reload")
-        return bool(self._get(url, timeout=self.timeout))
+        if resp := self._post(url, timeout=self.timeout):
+            logger.warning("reload: POST returned a non-empty response: %s", resp)
+            return False
+        return True
```
Abuelodelanada, line 40: If for a little moment Alertmanager API does not respond you'll get an error.
In `grafana-agent-operator` I implemented something [like this](https://github.com/canonical/grafana-agent-operator/blob/devel/src/charm.py#L314):

```python
    def _reload_config(self, attempts: int = 10) -> None:
        """Reload the config file.
        Args:
            attempts: number of attempts to reload
        """
        try:
            self.unit.status = MaintenanceStatus("reloading agent configuration")
            url = "http://localhost/-/reload"
            errors = list(range(400, 452)) + list(range(500, 513))
            s = Session()
            retries = Retry(total=attempts, backoff_factor=0.1, status_forcelist=errors)
            s.mount("http://", HTTPAdapter(max_retries=retries))
            s.post(url)
        except Exception as e:
            message = f"could not reload configuration: {str(e)}"
            raise GrafanaAgentReloadError(message)
```

src/alertmanager_client.py
```diff
@@ -33,21 +34,42 @@ def reload(self) -> bool:
...
+    def _get(url: str, timeout: float, data=None) -> bytes:
+        """Send a GET request.
+
+        The `data` arg must be None to tell urlopen it's a GET.
+        """
+        return Alertmanager._open(url, data, timeout)
+
+    @staticmethod
+    def _open(url: str, data: Optional[bytes], timeout: float) -> bytes:
+        """Send a request using urlopen.
```
Abuelodelanada, line 61: Are these methods necessary? Why not using the `request` module instead?

Abuelodelanada (approved):

PR #44, opened 2021-11-03
fix doc and type hints

sed-i:
- list out `tox` commands for running integration tests.
- appease mypy, at least partially, until types are [fixed in OF](https://github.com/canonical/operator/pull/664)
- use released python-libjuju from pypi instead of repo main (there was a regression recently that prompted this change)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -119,10 +119,11 @@ class AlertmanagerCharm(CharmBase):
         return self._api_port
 
     @property
-    def peer_relation(self) -> Optional[Relation]:
+    def peer_relation(self) -> Relation:
         """Helper function for obtaining the peer relation object.
 
-        Returns: peer relation object; returns None if called too early, e.g. during install.
+        Returns: peer relation object
+        (NOTE: would return None if called too early, e.g. during install).
         """
         return self.model.get_relation(self._peer_relation_name)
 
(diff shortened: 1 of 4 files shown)


Abuelodelanada (approved):
LGTM

src/charm.py
```diff
@@ -119,10 +119,11 @@ def api_port(self) -> int:
         return self._api_port
 
     @property
-    def peer_relation(self) -> Optional[Relation]:
+    def peer_relation(self) -> Relation:
```
mmanciop, line 122: well technically this could still return `None`, and afaiu this would still qualify for `Optional`

mmanciop (approved):

src/charm.py
sed-i: Correct. Totally agree, but in OF it's just Relation:
https://github.com/canonical/operator/blob/626527f988c007316218891830b1265fc936295e/ops/model.py#L137

```python
    def get_relation(
            self, relation_name: str,
            relation_id: typing.Optional[int] = None) -> 'Relation':
```
The question is what is correct and what should we do about it.
At the moment I think that OF should never return a None for a peer relation (type hint: Relation), but at the same time for other relation the type hint should be `Relation | None`.
This change here is a tradeoff. WDYT?

PR #45, opened 2021-11-07
add image sha to metadata

sed-i:
This PR introduces a version pinning for the default OCI image used by the charm.

- [x] specify sha for the upstream-source
- [x] use upstrem-source in integration tests when building charm locally
- [ ] document it in README.md

In the future this may also be handy for auto-publishing to charmhub.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -27,7 +27,7 @@ resources:
   alertmanager-image:
     type: oci-image
     description: OCI image for alertmanager
-    upstream-source: ubuntu/prometheus-alertmanager@sha256:30489547ce93a38e3cb3c76d102886e12a72145727f08ff690d2da0db697313d
+    upstream-source: ubuntu/prometheus-alertmanager:0.21-20.04_beta
 provides:
   alerting:
     # The provider (alertmanager) adds the following key-value pair to the relation data bag of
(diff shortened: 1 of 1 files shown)


metadata.yaml
```diff
@@ -27,7 +27,7 @@ resources:
   alertmanager-image:
     type: oci-image
     description: OCI image for alertmanager
-    upstream-source: ubuntu/prometheus-alertmanager
+    upstream-source: ubuntu/prometheus-alertmanager@sha256:30489547ce93a38e3cb3c76d102886e12a72145727f08ff690d2da0db697313d
```
sed-i, line 30: `sha256:304..13d` is for `amd64`:
https://hub.docker.com/layers/ubuntu/prometheus-alertmanager/0.21-20.04_beta/images/sha256-30489547ce93a38e3cb3c76d102886e12a72145727f08ff690d2da0db697313d?context=explore

Wouldn't it make more sense to use a tag here instead? Sure, a tag is not as trustworthy as a hash, but an arch-specific hash is probably the wrong way to go.

The docker hub url itself shows this clearly:
`hub.docker.com/.../0.21-20.04_beta/images/sha256-304...13d`
```suggestion
    upstream-source: ubuntu/prometheus-alertmanager:0.21-20.04_beta
```

WDYT? @simskij @mmanciop

Abuelodelanada (approved):

PR #46, opened 2021-11-07
mention amtool

sed-i:
Doc improvement: `curl -XPOST` is fine, but amtool is probably more convenient.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,15 @@ source .tox/unit/bin/activate
 ```
 
 #### Manual testing
-Alerts can be created using alertmanager's HTTP API,
+Alerts can be created using [`amtool`](https://manpages.debian.org/testing/prometheus-alertmanager/amtool.1.en.html),
+
+```shell
+amtool alert add alertname=oops service="my-service" severety=warning \
+    instance="oops.example.net" --annotation=summary="High latency is high!" \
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -73,7 +73,15 @@ source .tox/unit/bin/activate
 ```
 
 #### Manual testing
-Alerts can be created using alertmanager's HTTP API,
+Alerts can be created using [`amtool`](https://manpages.debian.org/testing/prometheus-alertmanager/amtool.1.en.html),
+
+```shell
+amtool alert add alertname=oops service="my-service" severety=warning \
```
mmanciop, line 79: The word “severety” is severely misspelled :D

Also the example uses addresses that do not sound quite realistic for LMA on K8s.

mmanciop (changes requested):

Abuelodelanada (approved):

mmanciop (approved):

CONTRIBUTING.md
sed-i: Fixed spelling.
Does it matter what address I give as an example for a manual test? What would you rather have there? Something like `prometheus.juju`?

mmanciop: I wish we knew how ingress routes will look like :-(

PR #50, opened 2021-12-16
enable upgrade test

sed-i:
`ops_test.model.applications[app_name].refresh` is still broken so resorting to cli

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from pathlib import Path
+
+import pytest
+from pytest_operator.plugin import OpsTest
+
+
+@pytest.fixture(scope="module")
+async def charm_under_test(ops_test: OpsTest) -> Path:
+    """Charm used for integration testing."""
+    charm = await ops_test.build_charm(".")
+    return charm
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -7,9 +7,14 @@ from pathlib import Path
 import pytest
 from pytest_operator.plugin import OpsTest
 
+_path_to_built_charm = None
+
 
 @pytest.fixture(scope="module")
 async def charm_under_test(ops_test: OpsTest) -> Path:
     """Charm used for integration testing."""
-    charm = await ops_test.build_charm(".")
-    return charm
+    global _path_to_built_charm
+    if _path_to_built_charm is None:
+        _path_to_built_charm = await ops_test.build_charm(".")
+
+    return _path_to_built_charm
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -7,14 +7,10 @@ from pathlib import Path
 import pytest
 from pytest_operator.plugin import OpsTest
 
-_path_to_built_charm = None
-
 
 @pytest.fixture(scope="module")
 async def charm_under_test(ops_test: OpsTest) -> Path:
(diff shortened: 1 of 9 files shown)


sed-i:
Action-master, aka @simskij, How do I prevent the `push` action running at the same time as `pull_request`?
https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662
Previously you proposed we do not specify "main" branch for "push" action.

rbarry82:
> Action-master, aka @simskij, How do I prevent the `push` action running at the same time as `pull_request`? https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662 Previously you proposed we do not specify only the "main" branch for "push" action.

Replace
```
name: Run-Tests
on: [pull_request, push]
```

With:
```
name: Run-Tests
on:
  push:
    branches:
      - '!main'
    # Or...
    branches-ignore: ['main']
  pull_request:
    branches:
      - main
 ```
 
 You can kind of filter this, but my parse from the question is "please run actions on every push to feature branches, and please also run them on PRs to `main`, but don't do both"

**Review by rbarry82 — APPROVED** [review]
I like everything about this patch

**Comment by sed-i** [comment]
> ```
> name: Run-Tests
> on:
>   push:
>     branches:
>       - '!main'
>     # Or...
>     branches-ignore: ['main']
>   pull_request:
>     branches:
>       - main
> ```

It looks like no push to main would trigger the wf.
What I have in mind is:
- every commit that is added to a PR is triggering a wf (but just once, not twice)
- every merge (push) to main triggers a wf

File: tests/integration/conftest.py [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+from pytest_operator.plugin import OpsTest
+
+_path_to_built_charm = None
+
+
+@pytest.fixture(scope="module")
+async def charm_under_test(ops_test: OpsTest) -> Path:
+    """Charm used for integration testing."""
+    global _path_to_built_charm
```
sed-i, line 16: @rbarry82 without this, the charm is being built again and again.
cf. [grafana](https://github.com/rbarry82/grafana-operator/blob/e9f169517988060d110e7fba87ac01b948f5c969/tests/integration/conftest.py#L25)

sed-i: https://github.com/charmed-kubernetes/pytest-operator/issues/40

rbarry82: The way I'm "cheating" it to build globals (without ugly `globals`) keywords in the integration test harness is actually with `pytest_configure` to find the "best" controller based on the projected workload (bundle mixed machine/sidecar, whatever).

`pytest_sessionstart` looks like it will do what you need, though.

There isn't an existing global event loop in this hook, but you can literally just:

```
def pytest_sessionstart():
    pytest.path_to_built_charm = asyncio.run(...)
```

simskij:
> It looks like no push to main would trigger the wf. What I have in mind is:
> [...]

```
name: Run-Tests
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

There is no way to do **exactly** what you suggest, but the above snippet will trigger on pushes to main (ie. post pr merge) and on PRs targeting main (including new pushes to those PRs). The trade-off being that it only runs on feature branches with an open pull-request, which I think is fine.

rbarry82 (approved):
Still +1

dstathis (approved):

simskij:
Btw, I think what I refered to as something not to do was to add them in this fashion:

```
name: Run-Tests
on: [pull_request, push]
```
precisely because of the multi-trigger issue. 😅

PR #51, opened 2021-12-17
add rescale test

sed-i:
Add a test for scale up/down and trigger leadership change on the way.

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -89,12 +89,13 @@ async def get_leader_unit_num(ops_test: OpsTest, app_name: str):
     return is_leader.index(True)
 
 
-async def block_until_leader_elected(ops_test: OpsTest, app_name: str):
-    async def is_leader_elected():
-        units = ops_test.model.applications[app_name].units
-        return any([await units[i].is_leader_from_status() for i in range(len(units))])
+async def is_leader_elected(ops_test, app_name):
+    units = ops_test.model.applications[app_name].units
+    return any([await units[i].is_leader_from_status() for i in range(len(units))])
 
+
+async def block_until_leader_elected(ops_test: OpsTest, app_name: str):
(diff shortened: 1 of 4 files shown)


simskij (approved):

tests/integration/helpers.py
```diff
@@ -79,3 +80,21 @@ async def __aenter__(self):
...
+
+async def block_until_leader_elected(ops_test: OpsTest, app_name: str):
+    async def is_leader_elected():
+        units = ops_test.model.applications[app_name].units
+        return any([await units[i].is_leader_from_status() for i in range(len(units))])
+
+    # await ops_test.model.block_until(is_leader_elected)
+    # block_until does not take async (yet?) https://github.com/juju/python-libjuju/issues/609
+    while not await is_leader_elected():
+        await asyncio.sleep(5)
```
Abuelodelanada, line 100: Why we are using inner function here?

tests/integration/test_rescale_charm.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+import yaml
+from helpers import (  # type: ignore[attr-defined]
+    IPAddressWorkaround,
+    block_until_leader_elected,
+    get_leader_unit_num,
+    get_unit_address,
+)
+
+log = logging.getLogger(__name__)
```
Abuelodelanada, line 28: Why not `logger` as usual?:

```python
```

Abuelodelanada (approved):

tests/integration/helpers.py
simskij: good catch! i thought we agreed not to do that :)

sed-i: Done

tests/integration/test_rescale_charm.py
sed-i: Done

PR #53, opened 2022-01-18
version pinning

### LINKED ISSUE #52: version freeze everything
Before promoting the alertmanager-k8s charm to beta it might be a good idea to version lock a few things:
 
- [ ] pip packages
- [ ] image versions
- [ ] add to metadata: assumes: juju >= 2.9.23
- [ ] juju version for ci itests:

```yaml
    - name: Setup operator environment
      uses: charmed-kubernetes/actions-operator@main
      with:
        juju-channel: latest/candidate
```

sed-i:
This PR introduces version pinning:
- requirements are now specified in `requirements.in`
- `tox -e pip-upgrade` would upgrade all production versions to the latest, subject to constraints in `requirements.in`

TODO:
- [X] pin prod reqs
- [X] add [`assumes`](https://discourse.charmhub.io/t/about-assumes/5450) for juju
- [ ] pin dev reqs

Fixes #52

References:

- https://pip-tools.readthedocs.io/en/latest/
- https://stackoverflow.com/a/36848206/3516684


simskij:
Would you mind expanding a bit on why this is preferable to a regular `pip freeze > requirements.txt`? Unless the benefits are **extremely** strong, I tend to lean toward keeping things simple with as few moving parts as possible.

sed-i:
> Would you mind expanding a bit on why this is preferable to a regular `pip freeze > requirements.txt`? Unless the benefits are **extremely** strong, I tend to lean toward keeping things simple with as few moving parts as possible.

The equivalent workflow with `pip freeze` is a bit awkward: dev env would pip install from requirements.in, but charmcraft would pip install from the frozen requirements.txt (need to remember to keep it up to date with requirements.txt).

sed-i:
Here's a manual `pip freeze` analogue (which can easily be adapted into a tox env).

```shell
virtualenv -p python3 venv_freeze
source venv_freeze/bin/activate
pip install -r requirements.in
pip freeze -r requirements.in > requirements.txt
```

Here is the diff between `pip-compile` (base) and `pip freeze` (diff) after stripping down comments:
```diff
anyio==3.5.0
certifi==2021.10.8
charset-normalizer==2.0.11
h11==0.12.0
httpcore==0.14.7
httpx==0.22.0
idna==3.3
lightkube==0.10.1
lightkube-models==1.23.1.4
ops @ https://github.com/canonical/operator/archive/refs/heads/main.zip
-pyyaml==6.0
+PyYAML==6.0
-rfc3986[idna2008]==1.5.0
+rfc3986==1.5.0
sniffio==1.2.0
```

`pip-compile` features (compared to `pip freeze`):
- "If pip-compile finds an existing requirements.txt file that fulfils the dependencies then no changes will be made, even if updates are available." When used as part of CI (i.e. "blindly") this is more consistent than pip freeze. `pip compile` would update an existing reqs file only if run with `--upgrade`, unlike `pip freeze`.
- (we do not have a use case for `pip-sync`)

sed-i:
Closing, per discussion.

PR #56, opened 2022-02-11
pip snapshot

sed-i:
A demo for creating pip freeze snapshot:
- call `tox -e pip-snapshot`
- the created `requirementes.snpashot` is just for reference and not used otherwise


sed-i:
After further discussing this it was decided to no freeze deps at this point.

PR #58, opened 2022-03-01
Use sighup to reload configuration

sed-i:
This PR uses sighup instead of the HTTP API [`POST /-/reload`](https://www.prometheus.io/docs/alerting/latest/management_api/#reload) endpoint to trigger config reload.

Unlike prometheus, alertmanager (0.23.0 / 2021-08-25 and earlier) does not have a `--web.enable-lifecycle` cli option, so nothing to do here on that front.

- There were a few surprising mypy errors for which I ended up doing fetch-lib so the diff looks bigger than what it really is.
- Renaming of `test_charm` to `test_config_changed_modifies_file` shows up as deletion + new file.

Crossref: OPENG-32

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,7 +17,7 @@ from ops.charm import ActionEvent, CharmBase
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, Relation
-from ops.pebble import Layer, PathError, ProtocolError
+from ops.pebble import APIError, Layer, PathError, ProtocolError
 
 from alertmanager_client import Alertmanager, AlertmanagerBadResponse
 
@@ -301,8 +301,9 @@ class AlertmanagerCharm(CharmBase):
         if config_hash == self._stored.config_hash:
             logger.debug("no change in config")
             return
+        else:
+            logger.debug("config changed")
 
-        logger.debug("config changed")
         self._push_config_and_reload(config_yaml)
         self._stored.config_hash = config_hash
 
@@ -334,12 +335,12 @@ class AlertmanagerCharm(CharmBase):
         except AlertmanagerBadResponse:
             config_from_server_before = None
 
-        # Send an HTTP POST to alertmanager to hot-reload the config.
+        # Send a SIGHUP to alertmanager to hot-reload the config.
         # This reduces down-time compared to restarting the service.
         try:
-            self.api.reload()
-        except AlertmanagerBadResponse as e:
-            logger.warning("config reload via HTTP POST failed: %s", str(e))
+            self.container.send_signal("SIGHUP", self._service_name)
+        except APIError as e:
+            logger.warning("config reload via SIGHUP failed: %s", str(e))
             # hot-reload failed so attempting a service restart
             if not self._restart_service():
                 raise ConfigUpdateFailure(


rbarry82 (approved):

balbirthomas:
Hi @sed-i @mmanciop @simskij 

- The HTTP API endpoint provides us with response status when we issue commands, which we lose when switching to using signals. 
- If the motivation of this PR is to  prevent external access of the HTTP (port 80) endpoint of the workload container inside a Alertmanager pod, shall we instead explore mechanisms in Juju or elsewhere that can help us do this ? Note the ports used for Prometheus and Alertmanager peers is not 80. Port 80 in the workload container only needs to be accessed by the charm side card container. Hence port 80 need not be exposed by the  pod.

mmanciop:
Isn't the port 80 _also_ the same we use to serve the requests from Karma over the [https://github.com/canonical/alertmanager-k8s-operator/blob/279f4bfc051c181f6ef91197320cedca1748c121/metadata.yaml#L36](`alertmanager_dispatch`) relation interface AND what powers the Alertmanager interface?

sed-i:
> is it really an option _not_ to expose it?

Depends on the topology you have in mind :) but yes, in general the HTTP API (of all cos apps) would need to be exposed.

rbarry82:
Grafana is also not port 80, but all of these expose HTTP endpoints which will come over ingress as 80 regardless, and be remapped. 

Unlike the others, Grafana _does not_ support reload method, and Grafana (in general) requires auth for any "dangerous" endpoints. Yes, they'll all need to be exposed, and filtering endpoints out is not realistic.

lib/charms/karma_k8s/v0/karma_dashboard.py
```diff
@@ -128,10 +131,37 @@ class RelationManagerBase(Object):
...
+                    "to be managed by this library, but instead it is '{}'".format(
+                        relation_name,
+                        INTERFACE_NAME,
+                        self.charm.meta.relations[relation_name].interface_name,
+                    )
+                )
+        except KeyError:
+            raise ValueError(
+                "Relation '{}' is not in the charm's metadata.yaml".format(relation_name)
+            )
```
pengale, line 163: There isn't a centralized way for charm libraries to claim that they own a specific name. If we adopt this approach as a pattern, we risk the situation where two libraries cannot be used in a charm simply because of a name collision, rather than any inherent incompatibilities between the libraries.

I think that I'd prefer to see libraries that take a relation name and interface name as an argument to a constructor, thus allowing charm authors to work around naming collisions. The constructors can use default names on those arguments for a "zero config" approach, but I think that it's important to avoid hard coding things like names in our libraries.

pengale (commented):
Had some comments about the validate method.

lib/charms/karma_k8s/v0/karma_dashboard.py
sed-i: We have similar things in [loki](https://github.com/canonical/loki-k8s-operator/blob/cb14570fd5de503aa9980423db2cd5f0499fa08b/lib/charms/loki_k8s/v0/loki_push_api.py#L1146), [grafana](https://github.com/canonical/grafana-k8s-operator/blob/5cc6999f64c2044f701a51b79e700dcd07c1bdde/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L742), [prometheus](https://github.com/canonical/prometheus-k8s-operator/blob/98ec8278fda13bf23091ffab18d00b23ebceeada/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L980), [alertmanager](https://github.com/canonical/alertmanager-k8s-operator/blob/279f4bfc051c181f6ef91197320cedca1748c121/lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py#L92).

Can you please further explain the collision scenario?
The only hard coded name here is `INTERFACE_NAME = "karma_dashboard` and that cannot be changed if a charm wants to relate to this particular karma.

To my understanding the only scenario this limits is reusing this library as is for other purposes.

Side note: relations, and not charms, should be 1st-class citizens (assuming there can be only one; otherwise both relations and charms can be 1st class citizens :) ).

sed-i:
After discussing this a bit it was decided to stick with HTTP API.

PR #60, opened 2022-03-16
add release automation

simskij:
Splits the CI into two workflows. `Release to edge` will use the `charming-actions` to release to edge on charmhub.

### DIFF
--- a/.github/workflows/release-edge.yaml
+++ b/.github/workflows/release-edge.yaml
@@ -1,0 +1,92 @@
+name: Release to Edge
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  lib-check:
+    name: Static analysis of /lib for Python 3.5
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up Python 3.5
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.5
+    - name: Install dependencies
+      run: python3 -m pip install tox
+    - name: Run static analysis for /lib for 3.5
+      run: tox -vve static-lib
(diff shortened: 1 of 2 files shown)


.github/workflows/release-edge.yaml
```diff
@@ -0,0 +1,92 @@
...
+          fetch-depth: 0
+      - name: Select charmhub channel
+        uses: canonical/charming-actions/channel@1.0.0
+        id: channel
+      - name: Upload charm to charmhub
+        uses: canonical/charming-actions/upload-charm@1.0.0
+        with:
+          credentials: "${{ secrets.CHARMHUB_TOKEN }}"
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          channel: "${{ steps.channel.outputs.name }}"
```
sed-i, line 92: Where is `edge` coming from? Is that a default for `canonical/charming-actions/channel@1.0.0`?

sed-i (approved):

.github/workflows/release-edge.yaml
mmanciop: Yes: https://github.com/canonical/charming-actions/blob/1bb3791dd3fe8343ea3a415788b747661e2da38c/upload-bundle/README.md#inputs

sed-i: Any reason not to be explicit about it in a file called `release-edge`?

simskij: Not really. In the next iteration ™️  we'll also publish PRs to charmhub branches, and pushes to `track/some_track_name` to `some_track_name/edge`, so it made sense to me to keep it calculated so that we may extend and adjust.

PR #62, opened 2022-03-17
Fetch-lib and related upkeep (salavging valuables from #58)

sed-i:
This PR is a bunch of valuable changes salvaged from #58:
- fetch-lib
- mypy updates due to fetch-lib
- smarter config update
- (Renaming of test_charm to test_config_changed_modifies_file shows up as deletion + new file.)


src/charm.py
```diff
@@ -298,27 +298,65 @@ def _update_config(self) -> None:
...
         config_hash = sha256(config_yaml)
 
-        logger.debug(
-            "config changed" if config_hash != self._stored.config_hash else "no change",
-        )
+        if config_hash == self._stored.config_hash:
+            logger.debug("no change in config")
+            return
+        else:
+            logger.debug("config changed")
```
Abuelodelanada, line 305: Because we have a `return`, the `else` is not necessary

Abuelodelanada (approved):

src/charm.py
sed-i: Removed. Thanks!

lib/charms/karma_k8s/v0/karma_dashboard.py
```diff
@@ -37,7 +37,10 @@ def __init__(self, *args):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
```
mmanciop, line 40: Current patch version on CharmHub is already 3, is this correct?

sed-i: Correct.

simskij: if it's already 3, shouldn't this be 4?

sed-i: This is the alertmanager charm, it does not own the karma lib.
This change is after `fetch-lib`, so yes, it matches charmhub.

balbirthomas (approved):

PR #63, opened 2022-03-23
Do not mock `get_system_info` in utests

sed-i:
`get_system_info` is now implemented in harness so there is no reason to mock it.

Also, removing this is needed to be able to use the upcoming https://github.com/canonical/operator/pull/724.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -19,7 +19,6 @@ class TestWithInitialHooks(unittest.TestCase):
 
     @patch_network_get(private_address="1.1.1.1")
     @patch.object(Alertmanager, "reload", tautology)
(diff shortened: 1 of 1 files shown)


rbarry82:
Looks ok to me, though we probably want to wait for that to merge, so I'll hold off on the +1

sed-i:
> Looks ok to me, though we probably want to wait for that to merge, so I'll hold off on the +1

@rbarry82 this change has nothing to do with the ops/724 PR (note how requirements.txt here is unchanged) -- `get_system_info` was already implemented a while ago.

jnsgruk (approved):

Abuelodelanada (approved):

PR #70, opened 2022-06-24
Adding self-monitoring, alert rules and grafana dashboard

### LINKED ISSUE #69: Add self-monitoring of Alertmanager metrics
### Enhancement Proposal

We should replicate in this charm something like https://github.com/canonical/prometheus-k8s-operator/pull/312

Abuelodelanada:
This PR fixes #69


src/charm.py
```diff
@@ -366,7 +376,10 @@ def _common_exit_hook(self) -> None:
             service := self.container.get_service(self._service_name)
         ) and service.is_running()
 
-        num_peers = len(self.peer_relation.units)
+        num_peers = 0
+
+        if self.peer_relation:
+            num_peers = len(self.peer_relation.units)
```
sed-i, line 382: Might be better if we call `get_relation` less times.
If you're in the mood for mini-golfing:

```suggestion
        num_peers = len(rel.units) if (rel := self.peer_relation) else 0
```

I forget - does rel.units include the querying unit? I.e. should we also have:
```python
num_peers = max(0, num_peers - 1)
```

tests/integration/test_update_status_pressure.py
```diff
@@ -42,12 +42,14 @@ async def test_deploy_multiple_units(ops_test: OpsTest, charm_under_test):
...
+        ops_test.model.deploy(
+            "ch:prometheus-k8s", application_name="prom", channel="edge", trust=True
+        ),
     )
 
     await asyncio.gather(
-        ops_test.model.add_relation(app_name, "prom"),
-        ops_test.model.wait_for_idle(status="active", timeout=300),
+        ops_test.model.add_relation(app_name, "prom:alertmanager"),
+        ops_test.model.wait_for_idle(status="active", timeout=2500),
```
sed-i, line 52: 2500 seems excessive. What kind of errors do you see with 300?

tests/integration/test_update_status_pressure.py
```diff
@@ -42,12 +42,14 @@ async def test_deploy_multiple_units(ops_test: OpsTest, charm_under_test):
...
-        ops_test.model.deploy("ch:prometheus-k8s", application_name="prom", channel="edge"),
+        ops_test.model.deploy(
+            "ch:prometheus-k8s", application_name="prom", channel="edge", trust=True
+        ),
     )
 
     await asyncio.gather(
-        ops_test.model.add_relation(app_name, "prom"),
-        ops_test.model.wait_for_idle(status="active", timeout=300),
+        ops_test.model.add_relation(app_name, "prom:alertmanager"),
```
sed-i, line 51: The test would be less fragile if we specify the relation name on `app_name` side instead of prom side.

tests/integration/test_update_status_pressure.py
Abuelodelanada: a lot of timeouts that had been solved with these increments.

src/charm.py
Abuelodelanada: No, it does not.

tests/integration/test_update_status_pressure.py
sed-i: I just noticed that this is the `update-status pressure` test, which sets `"update-status-hook-interval": "10s"`.
That's probably why you needed 2500, but I am surprised this helps at all: I'd expect it to never be idle long enough!

rbarry82 (approved):

PR #78, opened 2022-07-13
Integration test for persistence of silences

balbirthomas:
This PR adds an integration test for persistence of silences and hence also tests data persistence because silences are stored in the Alertmanager k8s charm's persistence volume claim. As part of the PR the Alertmanager client object is also enhanced to set, get and delete silences.

### DIFF
--- a/tests/integration/test_persistence.py
+++ b/tests/integration/test_persistence.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+import pytest
+import yaml
+from helpers import get_unit_address, is_alertmanager_up
+from pytest_operator.plugin import OpsTest
+
+from alertmanager_client import Alertmanager
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = METADATA["name"]
+resources = {"alertmanager-image": METADATA["resources"]["alertmanager-image"]["upstream-source"]}
+
+
+@pytest.mark.abort_on_fail
+async def test_silences_persist_across_upgrades(ops_test: OpsTest, charm_under_test, httpserver):
+    # deploy alertmanager charm from charmhub
+    logger.info("deploy charm from charmhub")
+    await ops_test.model.deploy("ch:alertmanager-k8s", application_name=app_name, channel="edge")
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+
+    # set a silencer for an alert and check it is set
+    alertmanager = Alertmanager(address=await get_unit_address(ops_test, app_name, 0))
+
+    silence_start = datetime.now(timezone.utc)
+    silence_end = silence_start + timedelta(minutes=30)
+    matchers = [
+        {
+            "name": "alertname",
+            "value": "fake-alert",
+            "isRegex": False,
+        }
+    ]
+    alertmanager.set_silences(matchers, silence_start, silence_end)
+    silences_before = alertmanager.get_silences()
+    assert len(silences_before)
+
+    # upgrade alertmanger using charm built locally
+    logger.info("upgrade deployed charm with local charm %s", charm_under_test)
+    await ops_test.model.applications[app_name].refresh(path=charm_under_test, resources=resources)
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+    assert await is_alertmanager_up(ops_test, app_name)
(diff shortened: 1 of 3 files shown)


tests/integration/test_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+            "value": "fake-alert",
+            "isRegex": False,
+        }
+    ]
+    alertmanager.set_silences(matchers, silence_start, silence_end)
+    silences = alertmanager.get_silences()
+    assert len(silences)
+
+    # Use kubectl to send a SIGTERM signal to Alertmanager so that data is flushed to disk.
+    # This step should not be necessary once the bug in the following issue ticket is fixed
```
sed-i, line 47:
```suggestion
    # This step should not be necessary once Juju 2.9.33 is released.
```

tests/integration/test_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+        container_name,
+        "--",
+    ]
+    # find pid of alertmanager
+    pid_cmd = ["pidof", "alertmanager"]
+    cmd = sg_cmd + [" ".join(kubectl_cmd + pid_cmd)]
+    retcode, alertmanager_pid, stderr = await ops_test.run(*cmd)
+    assert retcode == 0, f"kubectl failed: {(stderr or alertmanager_pid).strip()}"
+    # use pid of alertmanager to send it a SIGTERM signal using kubectl
+    term_cmd = ["kill", "-s", "TERM", alertmanager_pid]
```
sed-i, line 73: The following works for me:
```bash
juju exec --unit am/0 -- \
  PEBBLE_SOCKET=/charm/containers/alertmanager/pebble.socket \
  pebble signal TERM alertmanager
```
And if you restructure
```bash
juju exec --unit am/0 -- sh -c "PEBBLE_SOCKET=/charm/containers/alertmanager/pebble.socket pebble signal TERM alertmanager"
```
Then could probably use it in [`ops_test.juju("exec", ...)`](https://github.com/charmed-kubernetes/pytest-operator/blob/main/docs/reference.md#async-def-jujuself-args-kwargs).

tests/integration/test_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = METADATA["name"]
+resources = {"alertmanager-image": METADATA["resources"]["alertmanager-image"]["upstream-source"]}
+
+
+@pytest.mark.abort_on_fail
+async def test_silences_persist_across_upgrades(ops_test: OpsTest, charm_under_test, httpserver):
+    # deploy alertmanager charm from charmhub
+    logger.info("deploy charm from charmhub")
+    await ops_test.model.deploy("ch:alertmanager-k8s", application_name=app_name, channel="edge")
```
sed-i, line 27: Iiuc, for this test we can deploy the charm-under-test and then refresh with the same again, i.e. no need for a charmhub charm.

tests/integration/test_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+    # upgrade alertmanger using charm built locally
+    logger.info("upgrade deployed charm with local charm %s", charm_under_test)
+    await ops_test.model.applications[app_name].refresh(path=charm_under_test, resources=resources)
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+    assert await is_alertmanager_up(ops_test, app_name)
+
+    # check silencer is still set
+    alertmanager = Alertmanager(address=await get_unit_address(ops_test, app_name, 0))
+    silences = alertmanager.get_silences()
+    assert len(silences)
```
sed-i, line 92: Can we do here `assert silences_before == silences_after`?

tests/integration/test_persistence.py
balbirthomas: There is no guarantee that 2.9.33 is the release it will be fixed in, it could be earlier or later.

tests/integration/test_persistence.py
balbirthomas: I intentionally avoided the use of Pebble since the bug was related to pebble. If there is no strong reason to prefer it this way I am inclined to leave this temporary work around as is.

tests/integration/test_persistence.py
balbirthomas: Is there any strong reason to prefer doing it this way. I wanted to make the test as realistic as possible, in that the two versions of the charm being deployed are actually different. Besides by deploying from charmhub and upgrading from git we ensure detection if any future PR breaks persistence due to any incompatibility of currently released charm and proposed PR.

tests/integration/test_persistence.py
balbirthomas: Yes indeed, thank you. Done.

sed-i (approved):

PR #94, opened 2022-08-23
Consider ingress revoked in utests

sed-i:
Depends on https://github.com/canonical/traefik-k8s-operator/pull/84

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress.py
+++ b/lib/charms/traefik_k8s/v1/ingress.py
@@ -54,7 +54,7 @@ class SomeCharm(CharmBase):
 import logging
 import socket
 import typing
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Optional, Tuple
 
 import yaml
 from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent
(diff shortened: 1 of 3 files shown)


dstathis (approved):

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -493,15 +509,14 @@ def relation(self):
         """The established Relation instance, or None."""
         return self.relations[0] if self.relations else None
 
-    @property
-    def url(self) -> Optional[str]:
+    def _get_url_from_relation_data(self) -> str:
```
simskij, line 512: Same as in the other PR, should probably continue to return `None`.

simskij (changes requested):

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -511,11 +526,21 @@ def url(self) -> Optional[str]:
...
+
+    @property
+    def url(self) -> str:
+        """The full ingress URL to reach the current unit.
+
+        May return None if the URL isn't available yet.
+        """
+        data = self._stored.current_url or ""  # type: ignore
+        assert isinstance(data, str)  # for static checker
+        return data
```
Abuelodelanada, line 546: As in the other PR, shouldn't we change the return type here too?

sed-i: Updated.

PR #96, opened 2022-08-25
Upkeep

dstathis:
Issue:
Upkeep.

Solution:
Unpin dependencies, consolidate CI, and set workload_version.

Testing:
Deploy and check version is set.

Changelog:
alertmanager-k8s now sets workload version


src/charm.py
```diff
@@ -514,6 +514,11 @@ def _common_exit_hook(self) -> None:
     def _on_pebble_ready(self, _):
         """Event handler for PebbleReadyEvent."""
         self._common_exit_hook()
+        try:
+            version = self.api.version
+            self.unit.set_workload_version(version)
```
jnsgruk, line 519: I'd maybe add this into update_status too in case it fails here?

Abuelodelanada: +1 
We have no guarantee that when `pebble_ready` is fired, Alertmanager is up and running....
I would also add this to `upgrade_charm`, since there is a good chance that this led to an workload image upgrade.

Abuelodelanada (approved):

src/charm.py
dstathis: I agree with adding it to `update_status`

If the charm is upgraded and we get a new container, then `pebble_ready` should fire anyway, so we shouldn't need to do it in `upgrade_charm`.

sed-i: Consider moving `set_workload_version` outside the try-except.

sed-i (commented):
Finally a PR that deletes more than it adds! Woohoo!

src/charm.py
dstathis: @sed-i I had it in an else at first but this seemed more clear to me upon further consideration. It is identical functionally.

simskij: I'd add it to `upgrade_charm` as well to make the intent clear. It doesn't hurt, but also makes it a bit more reliable if Juju ever makes changes to the lifecycle.

src/charm.py
```diff
@@ -63,7 +64,7 @@ class AlertmanagerCharm(CharmBase):
     # Container name is automatically determined from charm name
     # Layer name is used for the layer label argument in container.add_layer
     # Service name matches charm name for consistency
-    _container_name = _layer_name = _service_name = "alertmanager"
+    _container_name = _layer_name = _service_name = _exe_name = "alertmanager"
```
sed-i, line 67: Why `exe` and not `bin`? Perhaps we talked too long about WSL :D
(Jk, this is _not_ a change request)

src/charm.py
```diff
@@ -629,6 +629,19 @@ def _external_url(self) -> str:
...
+    @property
+    def _alertmanager_version(self) -> str:
+        """Returns the version of alertmanager.
+
+        Returns:
+            A string equal to the agent version.
+        """
+        container = self.unit.get_container(self._container_name)
+        if not container.can_connect():
+            return None
```
sed-i, line 641: In that case the return type is `Optional[str]`. Odd if mypy didn't catch that.

src/charm.py
```diff
@@ -629,6 +629,19 @@ def _external_url(self) -> str:
...
+        version_output, _ = container.exec([self._exe_name, "--version"]).wait_output()
+        return re.search(r"version (\d*\.\d*\.\d*)", version_output).group(1)
```
sed-i, line 643: Here too mypy should have complained about a very possible None.

src/charm.py
dstathis: I was going to do _executable_name then decided it should be shorter 😁

src/charm.py
dstathis: I have changed the code to exec `alertmanager --version`. Thus we are not relying on the software actually being ready. So I think it is better not to rely on `update_status`.

I don't think running it in `upgrade_charm` makes any sense. If the lifecycle changes to the point where deploying a new image does not trigger pebble ready, then all of our charms will have much bigger problems than this.

src/charm.py
dstathis: Fixed.

src/charm.py
dstathis: Fixed.

src/charm.py
```diff
@@ -622,6 +629,24 @@ def _external_url(self) -> str:
...
+        container = self.unit.get_container(self._container_name)
+        if not container.can_connect():
+            return None
+        version_output, _ = container.exec([self._exe_name, "--version"]).wait_output()
+        # Output looks like this:
+        # alertmanager, version 0.23.0 (branch: HEAD, ...
+        result = re.search(r"version (\d*\.\d*\.\d*)", version_output)
+        if result is None:
+            return result
+        return result.group(1)
```
sed-i, line 648:
```suggestion
        result = re.search(r"version (\d*\.\d*\.\d*)", version_output)
        return result.group(1) if result else None
```
or
```suggestion
        if (result := re.search(r"version (\d*\.\d*\.\d*)", version_output)) is not None:
            return result.group(1)
        return None
```
or
```suggestion
        if (result := re.search(r"version (\d*\.\d*\.\d*)", version_output)) is not None:
            return result.group(1)
        # None is returned implicitly
```

tests/unit/helpers.py
```diff
@@ -33,3 +33,11 @@ def cli_arg(plan, cli_opt):
     _patch=tautology,
     is_ready=tautology,
 )
+
+
+class FakeProcessVersionCheck:
```
sed-i, line 38: Should we mock `_alertmanager_version` instead?

sed-i (approved):

tests/unit/helpers.py
dstathis: I sort of feed like it's best practice to mock the lowest level call. That way you exercise more of your code.

src/charm.py
rbarry82: The current version is :+1: from me. `.groupdict` is still more literate than the bad way Python handles result grouping from my POV, but I'm tainted from Perl dev anyway.

As a general comment (not that we're implementing any of them), can we please avoid walrus syntax wherever possible? We do it in the libraries, but as we speak of, say, a `grafana-agent` machine charm, it could probably _mostly_ share code with the existing codebase.

Following Python as they chase whatever syntax is cool these days (pattern matching, walrus, `functools....` a few years ago) saves a tiny bit of code, but also guarantees that it 💥 if we find a way to make a single `charm.py` usable in multiple cases. It's not worth the effort to strip it out yet, but we definitely don't need to be adding more outside of tests.

rbarry82 (approved):

PR #100, opened 2022-09-23
Clearing stale remote configuration data

Gmerold:
When validation of the remote Alertmanager configuration fails:
1. Clear relation data bag
2. Emit `configuration_broken` event, based on which Provider charm can be blocked

In such case, Alertmanager will fall back to the default dummy config.


dstathis (approved):
+1 Great change

simskij (approved):

PR #101, opened 2022-09-27
add catalogue integration

simskij:
This PR introduces a new relation interface called `catalogue` for connecting Alertmanager to a landing page. Over this relation, Alertmanager provides a catalogue item, which then will be used by the catalogue to display where users may go.

## Context
See https://github.com/canonical/catalogue-k8s-operator/ for the other end of this relation.

## Testing Instructions
- Build a charm file from this PR branch
- Deploy the build output on Juju
- Deploy `catalogue-k8s` from `edge`
- Relate the two charms together
- Visit the landing page on port 80

## Release Notes
Implement the `catalogue` interface to allow for integration with `catalogue-k8s`.


lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -233,22 +234,27 @@ class AlertmanagerProvider(RelationManagerBase):
...
     """
 
-    def __init__(self, charm, relation_name: str = "alerting", api_port: int = 9093):
+    def __init__(
+        self,
+        charm,
+        relation_name: str = "alerting",
+        api_port: int = 9093,
+        *,
+        external_url: Callable = None,
```
dstathis, line 243: Should we be using this external_url arg somewhere in the charm code?

dstathis, line 243: Shouldn't we be using the external_url arg somewhere in the charm code?

simskij: We should! Good catch!

sed-i (approved):

PR #103, opened 2022-10-13
fetch-lib

sed-i:
## Issue
Need to fetch lib prometheus_scrape to obtain ingress fixes.
Related to https://github.com/canonical/prometheus-k8s-operator/issues/368.

NOTE: depends on https://github.com/canonical/prometheus-k8s-operator/pull/384. Need to fetch-lib here again after it merges.

## Solution
- Pass `external_url` to MetricsEndpointProvider.
- Mind port 80 (ingress) when constructing the self-scrape job.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -140,6 +140,7 @@ class AlertmanagerCharm(CharmBase):
             self,
             relation_name="self-metrics-endpoint",
             jobs=[{"static_configs": [{"targets": [f"*:{self._ports.api}"]}]}],
+            external_url=self._external_url,
         )
 
         self.catalog = CatalogueConsumer(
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -191,6 +192,16 @@ def __init__(self, *args):
         self.framework.observe(self.on.show_config_action, self._on_show_config_action)
         self.framework.observe(self.on.check_config_action, self._on_check_config)
 
+    @property
+    def self_scraping_job(self):
+        """The self-monitoring scrape job."""
+        port = urlparse(self._external_url).port or 80
```
sed-i, line 198: @dstathis you probably want to do something similar with Loki:
https://github.com/canonical/loki-k8s-operator/blob/c95a7ae42b1b4585aaf485caa2a3e0b512475a0d/src/charm.py#L102

dstathis (approved):

sed-i:
@rbarry82 could you please publish-lib prometheus_scrape and fetch-lib here again?

PR #106, opened 2022-10-19
Update libraries

### LINKED ISSUE #105: Need to fetch-lib prometheus_scrape
#103 was merged before the latest prometheus_scrape was published (https://github.com/canonical/prometheus-k8s-operator/issues/384), so the issue described in https://github.com/canonical/prometheus-k8s-operator/issues/368 is probably still present for alertmanager.

Need to publish-lib prometheus_scrape and fetch-lib here.

rbarry82:
Basically `fetch-lib`. This is a marginal diff which is basically just `stripPrefix`, but either the rebuild of alertmanager from source or the new traefik corrects the catalogue index, so we may as well. 

Also `fetch-lib` for `prometheus_scrape`, which closes #105

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1,6 +1,8 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
-"""## Overview.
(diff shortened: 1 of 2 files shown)


rbarry82:
image
For example ^^

Abuelodelanada (approved):

PR #113, opened 2023-01-17
Fix RemoteConfigurationRequirer tests on ops 2.0.0

benhoyt:
Currently the TestAlertmanagerRemoteConfigurationRequirer unit tests fail on ops ops==2.0.0rc2 because begin_with_initial_hooks() has been changed to trigger pebble-ready.

You can try this by replacing "ops" in requirements.txt with "ops==2.0.0rc2" and running "tox -e unit", or just the specific tests:

```
tox -e unit -- -k TestAlertmanagerRemoteConfigurationRequirer
```

Once the Charm Tech team releases ops 2.0.0 final we can update requirements.txt to use it, and remove the TODO comment and set_can_connect() call.

### DIFF
--- a/tests/unit/test_remote_configuration_requirer.py
+++ b/tests/unit/test_remote_configuration_requirer.py
@@ -54,13 +54,15 @@ class TestAlertmanagerRemoteConfigurationRequirer(unittest.TestCase):
         # now does it.
         self.harness.set_can_connect("alertmanager", True)
 
-        with patch("ops.model.Container.exec") as patched_exec:
-            # In ops 2.0.0+, we need to mock exec, as begin_with_initial_hooks() now triggers
-            # pebble-ready, which execs "alertmanager --version".
-            patched_exec_mock = Mock()
-            patched_exec_mock.wait_output.return_value = ("alertmanager, version 0.23.0", "")
-            patched_exec.return_value = patched_exec_mock
-            self.harness.begin_with_initial_hooks()
+        # In ops 2.0.0+, we need to mock the version, as begin_with_initial_hooks() now triggers
+        # pebble-ready, which attempts to obtain the workload version.
+        patcher = patch.object(
+            AlertmanagerCharm, "_alertmanager_version", property(lambda *_: "0.0.0")
+        )
+        self.mock_version = patcher.start()
+        self.addCleanup(patcher.stop)
+
+        self.harness.begin_with_initial_hooks()
 
         self.relation_id = self.harness.add_relation(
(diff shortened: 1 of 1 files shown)


tests/unit/test_remote_configuration_requirer.py
```diff
@@ -49,8 +49,18 @@ def setUp(self, _) -> None:
...
+        # now does it.
         self.harness.set_can_connect("alertmanager", True)
-        self.harness.begin_with_initial_hooks()
+
+        with patch("ops.model.Container.exec") as patched_exec:
+            # In ops 2.0.0+, we need to mock exec, as begin_with_initial_hooks() now triggers
+            # pebble-ready, which execs "alertmanager --version".
+            patched_exec_mock = Mock()
+            patched_exec_mock.wait_output.return_value = ("alertmanager, version 0.23.0", "")
+            patched_exec.return_value = patched_exec_mock
```
sed-i, line 62: Should we have something like this in setUp?
```python
patcher = patch.object(AlertmanagerCharm, "_alertmanager_version", property(lambda *_: "0.0.0"))
self.mock_version = patcher.start()
self.addCleanup(patcher.stop)
```

sed-i (approved):
Thanks for the fix!

tests/unit/test_remote_configuration_requirer.py
benhoyt: Something like that would work. Would you prefer patching at the `_alertmanager_version` level rather than at the exec level? It's testing (or at least exercising) slightly more code patched at the exec level.

sed-i: Abstractions wise we could also get the version from the [web api](https://github.com/canonical/alertmanager-k8s-operator/blob/20b203d93931636101f22e11acb04220fb0e75e1/src/alertmanager_client.py#L117) so this mock wouldn't catch it.

(Ideally we would have a separate utest for the version parsing.)

Having a patcher would cover any future additions to this test class that may trigger the exec.

rbarry82: That's true, though any uncaught errors which may crash/cause a stack trace in the application itself may cause the workload version not to be fetched. We used to fetch this (not to set the workload version) way back in the initial days of OF charms, but the `_exec` is probably a nicer mechanism now, especially with go version output being very regular.

Nominally, I'd say that the patcher just on "side effect version call on startup" is just right, and other execs probably need different handling. A patcher for `_alertmanager_version` seems just right

benhoyt:
Thanks for the update -- ready to merge?

PR #114, opened 2023-01-17
Remove double observe

sed-i:
Ingress events were mistakenly observed twice.


src/charm.py
```diff
@@ -250,12 +243,6 @@ def _on_check_config(self, event: ActionEvent) -> None:
             {"result": output, "error-message": err, "valid": False if err else True}
         )
 
-    def _on_ingress_ready(self, event: IngressPerAppReadyEvent):
```
PietroPasotti, line 253: Does this charm do anything with the ingress address? Apparently not.

sed-i: It's used by `@property def _external_url(self)`.

PR #126, opened 2023-02-02
Run CI subject to ignorelist [just an idea - do not merge]

sed-i:
## Issue
When we make changes to e.g. `*.md` files, we don't need to run itests (or any CI currently) on pull request.
(Entire CI still runs on release-to-edge, which is good.)

TODO:
- [ ] Support globs

## Solution
- Create a `ci.ignorelist` for explicit, manual, opt-in for individual files.
- Run the `pull-request` CI only if the diff includes file(s) that are not listed in the ignorelist.

Alternative: https://github.com/marketplace/actions/get-diff-action


lucabello:
Minor aesthetic suggestion, could we rename the `ci.ignorelist` file to something like `.ciignore`? I like the consistency with `.gitignore`

lucabello:
Implementing in the shared CI with https://github.com/canonical/observability/pull/58.

PR #135, opened 2023-03-08
Find the uk8s group

rbarry82:
Don't always assume it's `microk8s`, it's not when strict.


sed-i:
Revision 54 of 'alertmanager-k8s' created

PR #136, opened 2023-03-10
Switch Alertmanager to use the ROCK on ghcr.io

simskij:
## Issue
We are currently running Alertmanager using the upstream OCI image.

## Solution
We are now building our own ROCKs for the observability stack. We should also make sure we're using them so that effort is not in vain.

## Testing Instructions
- Deploy the charm with the new resource
- Verify functionality is still there

## Release Notes
Switch from upstream OCI image to ROCK

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -34,7 +34,7 @@ resources:
   alertmanager-image:
     type: oci-image
     description: OCI image for alertmanager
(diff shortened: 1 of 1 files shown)


rbarry82 (approved):

Abuelodelanada (approved):

sed-i:
Was the rock rebuilt with version info? Pushed into the same tag?

PR #146, opened 2023-03-30
add heartbeat rule for integration with cos-alerter

### LINKED ISSUE #138: Create alert rule for cos-alerter
### Enhancement Proposal

We need an alert rule set up which will be always firing for cos-alerter.

dstathis:
Fixes #138 
Creates a continuously firing alert for use with cos-alerter. The severity is set to none so it should be okay to have even in the absence of cos-alerter.

### DIFF
--- a/src/prometheus_alert_rules/heartbeat.rule
+++ b/src/prometheus_alert_rules/heartbeat.rule
@@ -1,0 +1,7 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#prometheus-self-monitoring-1
+alert: Watchdog
+expr: vector(1)
+labels:
+  severity: none
+annotations:
+  summary: Continuously firing alert to ensure Alertmanager is working


src/prometheus_alert_rules/heartbeat.rule
```diff
@@ -0,0 +1,7 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#prometheus-self-monitoring-1
+alert: Watchdog
+expr: vector(1)
+labels:
+  severity: none
```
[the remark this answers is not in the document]
dstathis: I'm discarding it as of now. Won't prometheus_scrape add all of the juju topology anyway?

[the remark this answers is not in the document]
rbarry82: It will, at least, do that.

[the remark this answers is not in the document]
simskij: What prometheus scrape? This is an always-firing rule that won't be processed by any relation interfaces - unless I'm completely mistaken?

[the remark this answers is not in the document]
rbarry82: I mean, yes, but for it to fire at all:

* AM adds the rule here
* AM is related to Prometheus over prometheus_scrape
* The always firing rule is loaded+evaluated+starts firing
* Prom signals AM about an alert
* Off it goes for dispatch/checking

I think the assumption is that in order to make the always-firing rule be evaluated in the first place and start firing, it's going to Prom. If _all_ we want is to check Alertmanager itself, then we're back to "a secondary process which hits the AM API (or the addresses of every AM in the peer group/gossip ring and sends , or the ingress for AM, or whatever from somewhere else". 

If we're going to rely on an AlwaysFiring rule which will be sent from _one_ of the alertmanagers, and to trigger the DMS when all of them are gone, we need to actually fire the alert from somewhere for AM to route it in the first place.

src/prometheus_alert_rules/heartbeat.rule
```diff
@@ -0,0 +1,7 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#prometheus-self-monitoring-1
+alert: Watchdog
+expr: vector(1)
```
sed-i, line 3: **questions:** Do we know how an always-firing rule interacts with other alertmanager config? Should we document it here?

```
# ResolveTimeout is the default value used by alertmanager if the alert does
  # not include EndsAt, after this time passes it can declare the alert as resolved if it has not been updated.
  # This has no impact on alerts from Prometheus, as they always include EndsAt.
  [ resolve_timeout: [<duration>](https://prometheus.io/docs/alerting/latest/configuration/#duration) | default = 5m ]
```

```
# How long to initially wait to send a notification for a group
# of alerts. Allows to wait for an inhibiting alert to arrive or collect
# more initial alerts for the same group. (Usually ~0s to few minutes.)
[ group_wait: [<duration>](https://prometheus.io/docs/alerting/latest/configuration/#duration) | default = 30s ]
```

```
# How long to wait before sending a notification again if it has already
# been sent successfully for an alert. (Usually ~3h or more).
[ repeat_interval: [<duration>](https://prometheus.io/docs/alerting/latest/configuration/#duration) | default = 4h ]
```

src/prometheus_alert_rules/heartbeat.rule
```diff
@@ -0,0 +1,7 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#prometheus-self-monitoring-1
```
sed-i, line 1:
```suggestion
# Based on https://awesome-prometheus-alerts.grep.to/rules.html#prometheus-self-monitoring-1
# This is an always-firing alert for indicating if alertmanager itself is down (e.g. cos-alerter).
```

src/prometheus_alert_rules/heartbeat.rule
```diff
@@ -0,0 +1,7 @@
...
+alert: Watchdog
```
sed-i, line 2: **suggestion:** Name this so a causal look at alertmanager ui immediately discloses what it's for.

```suggestion
alert: AlertmanagerSelfWatchdog
```

src/prometheus_alert_rules/heartbeat.rule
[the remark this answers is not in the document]
dstathis: Yes this travels overprometheus_scrape to get from alertmanager-k8s to prometheus-k8s.

src/prometheus_alert_rules/heartbeat.rule
dstathis: Can you be more specific about what you want to know? As far as I can tell, those three values don't work any differently with this alert rule.

src/prometheus_alert_rules/heartbeat.rule
dstathis: I believe we have this in the summary already.

src/prometheus_alert_rules/heartbeat.rule
dstathis: I'm not sure `AlertmanagerSelfWatchdog` is any more clear. I don't think I would gain any info from that other than that it is a watchdog alert.

src/prometheus_alert_rules/heartbeat.rule
sed-i: I mean the "always firing" aspect:
- What does "always" mean on the receiver side?
- Could it start firing a bit too late just because alertmanager config set a high `group_wait`?
- Would it stop firing against our intention because `resolve_timeout` is set?
- Could we miss an alert if cos-alerter was down for a bit and missed the `repeat_interval`?
- Any potential negative interaction between `repeat_interval` here and the time duration conf'd in cos-alerter?

dstathis: * I'm not sure I understand. Always means this alert always resolves to firing.
* Yes group_wait should be `0s`. I'm adding a note in the cos-alerter readme.
* From the docs you posted: "This has no impact on alerts from Prometheus, as they always include EndsAt."
* This would be okay as presumably we have Prometheus configured to alert when cos-alerter goes down.
* Yes it is already documented in cos-alerter that `down_interval` (configured in cos-alerter) should be longer than `repeat_interval`

dstathis: I have no idea but I don't think we specify that in any alert rules right?

dstathis: [This issue](https://github.com/prometheus/prometheus/issues/5277) seems to state that it is long enough that it won't be an issue.

sed-i (approved):

PR #153, opened 2023-04-26
fire alert only once

### LINKED ISSUE #144: AlertmanagerNotificationsFailed will fire forever if fired once
With [`AlertmanagerNotificationsFailed`'s rule query](https://github.com/canonical/alertmanager-k8s-operator/blob/0138e87e8ce5dbf90af8437a30f94c317c7cd81b/src/prometheus_alert_rules/alertmanager_notifications_failed.rule#L2), once the alert fires once (i.e., there is at least a single notification failure), it'll remain fired till the end of times.

Something based on rate of changed or something similar is probably what's needed here.

lucabello:
Closes #144.

### DIFF
--- a/src/prometheus_alert_rules/alertmanager_notifications_failed.rule
+++ b/src/prometheus_alert_rules/alertmanager_notifications_failed.rule
@@ -1,5 +1,5 @@
 alert: AlertmanagerNotificationsFailed
-expr: alertmanager_notifications_failed_total{integration=~".*"} != 0
+expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0
 for: 0m
 labels:
   severity: warning


dstathis (approved):

src/prometheus_alert_rules/alertmanager_notifications_failed.rule
```diff
@@ -1,5 +1,5 @@
 alert: AlertmanagerNotificationsFailed
-expr: alertmanager_notifications_failed_total{integration=~".*"} != 0
+expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0
```
sed-i, line 2: **nitpick:** an `increase` would probably convey intent better.
Obviously, arithmetically the two are equivalent when the threshold is `0`.

PR #157, opened 2023-05-05
use pyright instead of mypy

lucabello:
Custom signals must be ignored in static checks with `# pyright: ignore` because they are only defined at runtime; **mypy** was just ignoring the entire **ops** module import, but that cannot be done with **pyright**.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,12 @@ per-file-ignores = {"tests/*" = ["D100","D101","D102","D103"]}
 convention = "google"
 
 # Static analysis tools configuration
+[tool.pyright]
+include = ["src"]
+extraPaths = ["lib"]
+pythonVersion = "3.8"
+pythonPlatform = "Linux"
+
 [tool.mypy]
 pretty = true
 python_version = 3.8
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,41 +29,9 @@ convention = "google"
 
 # Static analysis tools configuration
 [tool.pyright]
-include = ["src"]
 extraPaths = ["lib"]
 pythonVersion = "3.8"
 pythonPlatform = "Linux"
-
-[tool.mypy]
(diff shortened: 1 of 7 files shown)


Abuelodelanada (approved):

pyproject.toml
```diff
@@ -28,6 +28,12 @@ per-file-ignores = {"tests/*" = ["D100","D101","D102","D103"]}
 convention = "google"
 
 # Static analysis tools configuration
+[tool.pyright]
+include = ["src"]
+extraPaths = ["lib"]
```
sed-i, line 33: I wonder how this interacts with `charm,lib` commands below:
```
    charm: pyright {[vars]src_path} {posargs}
    lib: pyright --pythonversion 3.5 {[vars]lib_path} {posargs}
```
- If you pass `{[vars]src_path}`, `{[vars]lib_path}`, do we still need `include` and `extraPaths`?
- What's the difference between `include = ["src"]` and `include = ["src", "lib"]`?
- For the `lib` env, should `include` be `lib/charms/alertmanager-k8s` and `extraPaths` blank?
- Could/should these two be provided as args instead of config options?

src/charm.py
```diff
@@ -234,14 +242,14 @@ def _check_config(self) -> Tuple[str, str]:
         except ChangeError as e:
             output, err = "", e.err
         except ExecError as e:
-            output, err = e.stdout, e.stderr
+            output, err = str(e.stdout), str(e.stderr)
```
sed-i, line 245: Do we need `.decode().strip()` here? If it's coming directly from subprocess then stdout, stderr are bytes.

sed-i (approved):

src/charm.py
lucabello: That's why I'd assume this is fine :) If you agree please mark the conversation as resolved!

PR #160, opened 2023-05-15
add simple integration test to check cpu and memory config options

neilcam:
A simple integration test to verify the cpu and memory config options

### DIFF
--- a/tests/integration/test_resource_limits.py
+++ b/tests/integration/test_resource_limits.py
@@ -10,7 +10,7 @@ from pathlib import Path
 
 import pytest
 import yaml
-from helpers import  is_alertmanager_up
+from helpers import is_alertmanager_up
 from pytest_operator.plugin import OpsTest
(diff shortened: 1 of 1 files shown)


sed-i:
Thanks for this contribution @neilcam!
Compute resource limits is a functionality provided by a _charm library_ called [`kubernetes_compute_resources_patch`](https://charmhub.io/observability-libs/libraries/kubernetes_compute_resources_patch).
So I think the test included in this PR belongs in the [observability-libs repo](https://github.com/canonical/observability-libs/).
It possibly overlaps with the [existing tests](https://github.com/canonical/observability-libs/blob/main/tests/integration/test_compute_resources_patch.py).
Check it out?

neilcam:
Thanks for the comment! Had a feeling it might have been covered somewhere but didn't spot it in observability-libs tests! Thanks!

PR #168, opened 2023-06-29
PoC: generic workload manager

lucabello:


src/alertmanager.py
```diff
@@ -0,0 +1,179 @@
...
+
+from alertmanager_client import Alertmanager, AlertmanagerBadResponse
+from ops.pebble import ExecError, Layer
+
+from src.workload_manager import ConfigUpdateFailure, ContainerNotReady, WorkloadManager
+
+logger = logging.getLogger(__name__)
+
+
+class AlertmanagerWorkloadManager(WorkloadManager):
```
PietroPasotti, line 19:
```suggestion
class AlertmanagerManager(ManagerManager[Manager]):
```

src/alertmanager.py
```diff
@@ -0,0 +1,179 @@
...
+            container_name=container_name,
+            external_url=external_url,
+            config_path=config_path,
+            tls_enabled=tls_enabled,
+        )
+
+        self._peer_addresses = peer_addresses
+        self._api_port = api_port
+        self._ha_port = ha_port
+        self.api = Alertmanager(port=self._api_port, web_route_prefix=web_route_prefix)
```
PietroPasotti, line 50: proposal: rename 
Alertmanager -> Api
AlertmanagerWorkloadManager -> Workload
the module name shouldn't be repeated in all classes defined in it.

src/workload_manager.py
```diff
@@ -116,92 +103,35 @@ def is_ready(self):
...
-        # Output looks like this:
-        # alertmanager, version 0.23.0 (branch: HEAD, ...
-        result = re.search(r"version (\d*\.\d*\.\d*)", version_output)
-        if result is None:
-            return result
-        return result.group(1)
+        pass
 
+    @abstractmethod
     def check_config(self) -> Tuple[str, str]:
```
PietroPasotti, line 124: I'm not sure how well this will generalize.
Why would a 'check_config' method return stderr, stdout? I'd expect it to raise if invalid, or return bool.
This is very specific to the action we use it for, and the existence of amtool. 
Maybe all our charms have such a tool, however?

src/workload_manager.py
```diff
@@ -116,92 +103,35 @@ def is_ready(self):
...
-                        "override": "replace",
-                        "summary": "alertmanager service",
-                        "command": _command(),
-                        "startup": "enabled",
-                    }
-                },
-            }
-        )
+    @abstractmethod
+    def _workload_layer(self) -> Layer:
```
PietroPasotti, line 132:
```suggestion
    def _generate_layer(self) -> Layer:
```

src/workload_manager.py
```diff
@@ -248,7 +175,7 @@ def update_config(self, manifest: ConfigFileSystemState) -> None:
         logger.debug("applying config changes")
         manifest.apply(self._container)
 
-        # Validate with amtool and raise if bad
+        # Validate and raise if bad
         try:
             self.check_config()
         except WorkloadManagerError as e:
```
PietroPasotti, line 181: this is not true though. in the action, we set failed status if stdout is empty and stderr is nonempty. Should we raise here on those conditions too? Or (probably a bit more general): raise a custom exception from check_config, store stderr, stdout in it, and pull them out again to set the action results

PietroPasotti:
all in all I quite like where this is going, but I'm not sure we've nailed the abstractions yet.
We should make sure we have the right types and that each type encapsulates something real

src/workload_manager.py
sed-i, line 1: Much of this is still alertmanager-specific:
- Constructor args take `config_path` (what if I have more? or none?) and `tls_enabled` (why not cert?)
- `update_layer`, `restart_service` assume only one service.

Seems to me that an abstract class is not appropriate here because all we have is abstract methods, and there is only one derived class - AlertmanagerWorkload. I.e. we don't have a collection of different managers that we iterate and want to invoke the same method for all (=polymorphism).

Seems to me that a [Protocol](https://peps.python.org/pep-0544/) is also not appropriate because why should we force some interface across different charms (workloads)? If we want to be helpful across charms, we should offer functionality, not some "draconian" contract (=interface).

This reminds me how @PietroPasotti contrsted framework with toolkit: having a general purpose `workload_manager` is a toolkit-like ambition, but after stripping down, what's left is framework-like.

lucabello:
We concluded that the code was indeed too specific to generalize; closing the PR!

PR #171, opened 2023-07-07
Enable TLS for `alertmanager_dispatch` and the self-monitoring scrape job

### LINKED ISSUE #165: TLS part 2: Enable TLS for Prometheus alerting and self-monitoring
Tracking issue: canonical/cos-lite-bundle#75 

canonical/alertmanager-k8s-operator#164 enabled TLS for Alertmanager but both scrape target and `cluster_info` are still HTTP.

Potentially blocked by https://github.com/canonical/tls-certificates-interface/issues/46 (prom may not be able to scrape).

Related: https://github.com/canonical/charm-microk8s/pull/88

TODO:
- [ ] Update the self-scrape config to use https scheme
- [ ] Update the `cluster_info` logic in the `alerting` lib so prometheus charm generates the `alertmanagers` section correctly.
- [ ] Think what needs to be changed so that prometheus WITHOUT tls is able to scrape alertmanager WITH tls enabled (probably need to set insecure-skip-verify somewhere in prometheus).

sed-i:
## Problem
The `alertmanagers` section in prometheus's config is rendered without a scheme and without `tls_config`, and prometheus uses `http` by default.

## Solution
- [x] Add scheme to the "alerting" relation data
- [x] Add scheme for the self-monitoring scrape job
- [ ] ~~Simplify charm by using `strip-prefix` middleware~~

This PR goes in tandem with:
- https://github.com/canonical/prometheus-k8s-operator/pull/495
- https://github.com/canonical/observability-libs/pull/52

Reference: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tls_config

Fixes #165.


tests/scenario/test_brute_isolated.py
sed-i, line 1: @PietroPasotti could you please take a look at these scenario tests?
- Added some TODO with open questions
- After reshaping the tests scope etc. perhaps a different structure would make more sense.

Feel free to commit suggestions!

PietroPasotti: I've taken a look and I agree that there's something fishy about this but I can't quite put my finger on it yet.
I think the main thing is I'm used to writing scenario tests like so:
- arrange: create a State and a Context
- act: context.run(event, state)
- assert: check output state or side effects

and usually I can check everything I need to check based on this simple structure.
Chaining multiple `state = context.run(event, state)` calls in a single test feels odd.

IMHO, if you have:
```python
state0 = State(...)
state1 = context.run(event1, state0)
state2 = context.run(event2, state1)
state3 = context.run(event3, state2)
```

then you should be able to write
```python
state4 = State(...)  
state5 = context.run(event3, state4)
```

and the `state5` you obtain should be equal to `state3` you obtained in that multi-stage manner. If not, there is something weird about your charm (it's not stateless!).
Crux of the matter is of course how you construct state4, and I agree that there's some tension between hardcoding in state4 a bunch of stuff that could just as well be derived 'dynamically' by executing the sequence of events that led to that state. 

Let me try out something else.

tests/scenario/test_transition_to_https.py
```diff
@@ -0,0 +1,129 @@
...
+def initial_state(context, is_leader):
+    """An isolated alertmanager instance after the startup sequence."""
+    return begin_with_initial_hooks_isolated(context, leader=is_leader)
+
+
+@pytest.fixture
+def state_with_alerting_relation(initial_state, context, is_leader):
+    """Alertmanager after an alerting relation has been added."""
+    return add_relation_sequence(context, initial_state, Relation("alerting", relation_id=10))
+
```
sed-i, line 70: With all these fixtures in place it seems a bit too deconstructed.
I find it hard to reason about the tests with such a deconstruction.

tests/scenario/test_server_scheme.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+# https://stackoverflow.com/a/62176555/3516684
+
+
+# | Gherkin element | pytest artifact |
+# |-----------------|-----------------|
+# | Feature         | *.py            |
+# | Scenario        | Class           |
+# | Step            | Method          |
```
Abuelodelanada, line 22:
```suggestion
```

Abuelodelanada (approved):

tests/scenario/test_brute_isolated.py
sed-i: Thanks @PietroPasotti. 
Let's revisit this a bit later.

PR #180, opened 2023-08-11
add proxy config from env variables

### LINKED ISSUE #178: Support proxies
Looks like this should just be setting the http_proxy environment variable for Grafana based on a config value. We should likely do this for Alertmanager as well

_Originally posted by @dstathis in https://github.com/canonical/grafana-k8s-operator/issues/240#issuecomment-1668025553_

lucabello:
## Issues
Closes #178.

## Solution
Similarly to Grafana, we're honoring the proxy configuration set in the juju model config.

**Additional changes**: `charmcraft fetch-lib` :)


sed-i:
This was added as part of #195. Closing.

PR #182, opened 2023-08-14
Deprecate "web_external_url"; use strip_prefix

sed-i:
This PR deprecates the `web_external_url` config option in favor of strip_prefix.

Drive-by fixes:
- Use `open_port` instead of the k8s service patch.

---

[Y-Statement](https://adr.github.io/): In the context of end-to-end TLS, facing growing complexity in tracking URL path prefix, we decided to use `strip_prefix` and deprecate `web_external_url` to achieve simpler design, accepting the downside of requiring ingress for external reachability.


src/charm.py
```diff
@@ -127,13 +128,8 @@ def __init__(self, *args):
...
         self.remote_configuration = RemoteConfigurationRequirer(self)
 
-        api = ServicePort(self._ports.api, name=f"{self.app.name}")
-        ha = ServicePort(self._ports.ha, name=f"{self.app.name}-ha")
-        self.service_patcher = KubernetesServicePatch(
-            charm=self,
-            service_type="ClusterIP",
-            ports=[api, ha],
-        )
+        self.set_ports()
```
Abuelodelanada, line 131: Shouldn't we catch a possibly `ModelError` exception raised by `open_port` or `close_port`?

Abuelodelanada (approved):

src/charm.py
Abuelodelanada: We can let the charm goes into `ErrorStatus`, there is no problem with that.

My question is:  Before that, shouldn't we "unset" the ports that were setted?

Perhaps I am completely wrong, but seems like `set_ports()` should be a transactional action.

sed-i: > shouldn't we "unset" the ports that were setted?

We could, but I imagine in this case we'd rather have "degraded mode" than nonoperational.
It's also "corrected" every hook.

PR #195, opened 2023-09-26
Always forward to prometheus the fqdn URL, not ingress

### LINKED ISSUE #194: Use fqdn and correct scheme in `alertmanager_dispatch`
Currently we have two problems with the address `alertmanager_dispatch` is forwarding to prometheus:

https://github.com/canonical/alertmanager-k8s-operator/blob/99dc29b4cd1e4d5c345b5c8967d6ca858e601f21/lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py#L291

1. It is always "http" for fqdn.
2. It might be ingress address, whereas we only want fqdn for now. When one of the following is impl'd, we can go back to potentially forwarding an ingress address:
    - https://github.com/canonical/prometheus-k8s-operator/issues/530 
    - https://github.com/canonical/operator/issues/970

sed-i:
## Problem
`alertmanager_dispatch` forwards ingress url to prometheus, but prometheus may not have a cert in place to trust the ca that signed the ingress provider (traefik).

## Solution
Forward the k8s fqdn, until https://github.com/canonical/operator/issues/970 is impl'd.

Fixes #194.

Drive-by fixes:
- Improve provider API, and bump LIBAPI to 1.
- Incorporate proxy envvars from #180.

## Testing
Deploy the manual testing bundle, [bundle_1_e2e_tls.yaml](https://github.com/canonical/alertmanager-k8s-operator/blob/main/tests/manual/bundle_1_e2e_tls.yaml), and make sure prometheus was able to post alerts to alertmanager:
```
$ curl -k https://10.1.166.69:9093/api/v2/alerts
[{"annotations":{"description":"test_09f2c30f- ... and so it goes
```

### DIFF
--- a/lib/charms/alertmanager_k8s/v1/alertmanager_dispatch.py
+++ b/lib/charms/alertmanager_k8s/v1/alertmanager_dispatch.py
@@ -233,12 +233,18 @@ class AlertmanagerConsumer(RelationManagerBase):
         alertmanagers: Set[str] = set()
         for unit in relation.units:
             if rel_data := relation.data[unit]:
-                try:
-                    data = _ProviderSchemaV0(**rel_data)
-                except pydantic.ValidationError as e:
-                    logger.warning("Relation data failed validation: %s", e)
+                try:  # v1
+                    data = _ProviderSchemaV1(**rel_data)
+                except pydantic.ValidationError as ev1:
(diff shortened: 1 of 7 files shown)


Abuelodelanada (approved):
Manually tested!

PR #199, opened 2023-10-03
Fix catalogue code ordering issue

sed-i:
This PR moves the catalogue item from the constructor to the common exit hook.

In tandem with:
- https://github.com/canonical/catalogue-k8s-operator/issues/42

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -148,6 +148,27 @@ class AlertmanagerCharm(CharmBase):
             ],
         )
 
+        self.catalog = CatalogueConsumer(
+            charm=self,
+            refresh_event=[
+                self.ingress.on.ready,  # pyright: ignore
+                self.ingress.on.revoked,  # pyright: ignore
+                self.on["ingress"].relation_changed,
+                self.on.update_status,
+                self.on.config_changed,  # also covers upgrade-charm
+            ],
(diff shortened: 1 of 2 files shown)


Abuelodelanada (approved):

src/charm.py
```diff
@@ -450,6 +429,24 @@ def _common_exit_hook(self, update_ca_certs: bool = False) -> None:
...
+                icon="bell-alert",
+                url=self._external_url,
+                description=(
+                    "Alertmanager receives alerts from supporting applications, such as "
+                    "Prometheus or Loki, then deduplicates, groups and routes them to "
+                    "the configured receiver(s)."
+                ),
+            ),
+        )
+
```
simskij, line 449: If reloading alertmanager fails, the item won't show up in the catalogue. Is that intentional? It seems like we'd want to list broken items as well, so that it is easy to discover that they indeed are broken.

sed-i: Only if the common exit hook _never_ ran to the end then we won't have a catalogue entry.
But that also means we'd have a bad status.

I could instead use the new lib from https://github.com/canonical/catalogue-k8s-operator/pull/42.

sed-i: Changed to use the update method from catalog v1.

PR #200, opened 2023-10-05
use major version for rock tag

lucabello:
We should set the base image to track the major version tag for the ROCK; on a charm release, it will be fixed to the digest of the current latest image with the same major version.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -40,7 +40,7 @@ resources:
       OCI image for alertmanager. This charms makes the following assumptions about the image:
       - location of executable "alertmanager" is in the path
       - has `update-ca-certificates`
-    upstream-source: ubuntu/alertmanager:0.25.0-22.04_stable
+    upstream-source: ubuntu/alertmanager:0-22.04
 
 provides:
   alerting:


metadata.yaml
```diff
@@ -40,7 +40,7 @@ resources:
       OCI image for alertmanager. This charms makes the following assumptions about the image:
       - location of executable "alertmanager" is in the path
       - has `update-ca-certificates`
-    upstream-source: ubuntu/alertmanager:0.25.0-22.04_stable
+    upstream-source: ubuntu/alertmanager:0-22.04
```
sed-i, line 43: At first glance I thought the version tag is wrong because of the odd zero :)

For workloads that are in the zero major version, I think we should go down to minor version.
We do not want big breaking changes going unnoticed.

```suggestion
    upstream-source: ubuntu/alertmanager:0.25-22.04
```

sed-i (changes requested):

metadata.yaml
lucabello: Big breaking changes don't happen on minor versions; tracking `0` means that the charm will use `0.25`, `0.26`, etc; if `1.0` comes out, it won't be a `0.x` version, so it won't be tracked by this tag.

We discussed this already multiple times, but we agreed that jumping between minor versions is fine :)

simskij (approved):

PR #202, opened 2023-10-13
Use collect-status

sed-i:
## Issue
Status updates are communicated via exceptions.

## Solution
Use [collect-status](https://github.com/canonical/operator/pull/954).

In tandem with:
- https://github.com/canonical/observability-libs/pull/63

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -54,6 +54,17 @@ logger = logging.getLogger(__name__)
 PORTS = {"api": Port("tcp", 9093), "ha": Port("tcp", 9094)}
 
 
+def _is_external_url_valid(url: Optional[str]) -> bool:
+    if not url:
+        return False
+
+    parsed = urlparse(url)
+    if not (parsed.scheme in ["http", "https"] and parsed.hostname):
+        return False
+
+    return True
+
+
 class AlertmanagerCharm(CharmBase):
     """A Juju charm for alertmanager.
 
@@ -61,6 +72,8 @@ class AlertmanagerCharm(CharmBase):
         api: an API client instance for communicating with the alertmanager workload
                 server
     """
+
+    last_error: str
 
     # Container name must match metadata.yaml
     # Layer name is used for the layer label argument in container.add_layer
@@ -84,6 +97,8 @@ class AlertmanagerCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
         self.container = self.unit.get_container(self._container_name)
+
+        self.last_error: Optional[str] = None
 
         self.server_cert = CertHandler(
             self,
@@ -207,9 +222,24 @@ class AlertmanagerCharm(CharmBase):
                 MaintenanceStatus(f"Waiting for '{self.container.name}' container to become ready")
             )
 
+        # Make sure the external url is valid
+        if not _is_external_url_valid(self._external_url):
+            # This shouldn't happen
+            event.add_status(
+                BlockedStatus(
+                    f"Invalid external url: '{self._external_url}'; must include scheme and hostname."
+                )
+            )
+
         # TODO move this check into CertHandler?
         if self.server_cert.enabled and not self._is_tls_ready():
             event.add_status(MaintenanceStatus("Waiting for TLS ready"))
+
+        if self.last_error:
+            event.add_status(BlockedStatus(self.last_error))
+
+        # Must add ActiveStatus, otherwise charm ends up in UnknownStatus
+        event.add_status(ActiveStatus())
 
     @property
     def _catalogue_item(self) -> CatalogueItem:
@@ -373,27 +403,19 @@ class AlertmanagerCharm(CharmBase):
     def _common_exit_hook(self, update_ca_certs: bool = False) -> None:
         """Event processing hook that is common to all events to ensure idempotency."""
         if not self.resources_patch.is_ready():
-            if isinstance(self.unit.status, ActiveStatus) or self.unit.status.message == "":
-                self.unit.status = WaitingStatus("Waiting for resource limit patch to apply")
             return
 
         if not self.container.can_connect():
-            self.unit.status = MaintenanceStatus("Waiting for pod startup to complete")
             return
 
         # Make sure the external url is valid
-        if external_url := self._external_url:
-            parsed = urlparse(external_url)
-            if not (parsed.scheme in ["http", "https"] and parsed.hostname):
-                # This shouldn't happen
-                logger.error(
-                    "Invalid external url: '%s'; must include scheme and hostname.",
-                    external_url,
-                )
-                self.unit.status = BlockedStatus(
-                    f"Invalid external url: '{external_url}'; must include scheme and hostname."
-                )
-                return
+        if not _is_external_url_valid(self._external_url):
+            # This shouldn't happen
+            logger.error(
+                "Invalid external url: '%s'; must include scheme and hostname.",
+                self._external_url,
+            )
+            return
 
         # TODO Conditionally update with the external URL if it's a CMR, or rely on "recv-ca-cert"
         #  on the prometheus side.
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
+            )
+        if self.resources_patch.last_error:
+            event.add_status(BlockedStatus(f"[resource patch] {self.last_error}"))
 
-        new_ports_to_open = planned_ports.difference(actual_ports)
-        for p in new_ports_to_open:
-            self.unit.open_port(p.protocol, p.port)
+        # Check open ports
+        if self.unit.opened_ports() != set(PORTS.values()):
+            event.add_status(WaitingStatus(f'Opening {", ".join(PORTS.keys())} ports'))
```
dstathis, line 223: This should be maintenance status. Waiting is for waiting on related charms.

src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
+            )
+
+        # Make sure the external url is valid
+        if not _is_external_url_valid(self._external_url):
+            # This shouldn't happen
+            event.add_status(
+                BlockedStatus(
+                    f"Invalid external url: '{self._external_url}'; must include scheme and hostname."
+                )
+            )
```
dstathis, line 238: I believe this should be an error status. The external url is not set manually right? So there is no admin intervention that can fix it. The fact that you say "This shouldn't happen" is likely a hint that it should be error as well.

src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
+
+        # TLS
+        if self.server_cert.enabled and not self._is_tls_ready():
+            event.add_status(MaintenanceStatus("Waiting for TLS ready"))
+
+        if self.last_error:
+            event.add_status(BlockedStatus(self.last_error))
```
dstathis, line 245: I think this should also be error status I think because these, as far as I can tell, are miscellaneous un-handled errors.

src/charm.py
```diff
@@ -383,27 +409,19 @@ def _render_manifest(self) -> ConfigFileSystemState:
...
-                    f"Invalid external url: '{external_url}'; must include scheme and hostname."
-                )
-                return
+        if not _is_external_url_valid(self._external_url):
+            # This shouldn't happen
+            logger.error(
+                "Invalid external url: '%s'; must include scheme and hostname.",
+                self._external_url,
+            )
+            return
```
dstathis, line 424: Set `self.last_error` here?

tests/unit/test_charm.py
```diff
@@ -113,10 +113,16 @@ def test_topology_added_if_user_provided_config_with_group_by(self, *unused):
     def test_charm_blocks_if_user_provided_config_with_templates(self, *unused):
         new_config = yaml.dump({"templates": ["/what/ever/*.tmpl"]})
         self.harness.update_config({"config_file": new_config})
+        self.harness.evaluate_status()
         self.assertIsInstance(self.harness.charm.unit.status, BlockedStatus)
 
+        # Manual cleanup, due to https://github.com/canonical/operator/issues/736
+        self.harness.charm.unit._collected_statuses.clear()
```
benhoyt, line 120: Presumably this part of it (and similar lines below) will be fixed by https://github.com/canonical/operator/pull/1048?

src/charm.py
```diff
@@ -29,28 +29,41 @@
...
-    OpenedPort,
+    Port,
     Relation,
     WaitingStatus,
 )
 from ops.pebble import PathError, ProtocolError  # type: ignore
 
 logger = logging.getLogger(__name__)
 
+PORTS = {"api": Port("tcp", 9093), "ha": Port("tcp", 9094)}
```
benhoyt, line 54: Per other discussion, I think you'd be better off with:

```python
API_PORT = 9093
HA_PORT = 9094
PORTS = [API_PORT, HA_PORT]
...
self.unit.set_ports(*PORTS)
```

src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
 
-        # Ports may change across an upgrade, so need to sync
-        ports_to_close = actual_ports.difference(planned_ports)
-        for p in ports_to_close:
-            self.unit.close_port(p.protocol, p.port)
+    def _on_collect_unit_status(self, event: CollectStatusEvent):
+        # Resource limit
+        if not self.resources_patch.is_ready():
+            event.add_status(
+                MaintenanceStatus("[resource patch] Waiting for resource limit patch to apply")
```
benhoyt, line 216: When designing collect-status, we thought about including a prefix/component name automatically, but decided against it as status messages should be nicely-designed UI for the user. This kind of shows why: the fact that it's resource patch-related is already clear from the message part, so the `[resource patch]` prefix just adds noise to the UI. Something to consider...

src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
+            )
+        if self.resources_patch.last_error:
+            event.add_status(BlockedStatus(f"[resource patch] {self.last_error}"))
```
benhoyt, line 219: Should this be:
```suggestion
            event.add_status(BlockedStatus(f"[resource patch] {self.resources_patch.last_error}"))
```

benhoyt: Also, I can see why you used the prefix here, however, will the error messages from the resource patch object already include enough context to avoid the need for it, and give a nicer status message?

src/charm.py
benhoyt: True. Although I wonder whether this particular check should probably be removed entirely. `open-port` goes through right away. Juju itself may take time to perform it, but the charm doesn't, and this status won't reflect when the K8s operation actually goes through anyway.

src/charm.py
```diff
@@ -196,22 +207,45 @@ def __init__(self, *args):
...
-        for p in new_ports_to_open:
-            self.unit.open_port(p.protocol, p.port)
+        # Check open ports
+        if self.unit.opened_ports() != set(PORTS.values()):
+            event.add_status(WaitingStatus(f'Opening {", ".join(PORTS.keys())} ports'))
+
+        # "pebble ready"
+        if not self.container.can_connect():
+            event.add_status(
+                MaintenanceStatus(f"Waiting for '{self.container.name}' container to become ready")
```
benhoyt, line 228: Consider the following. "to become ready" is kind of redundant.

```suggestion
                MaintenanceStatus(f"Waiting for container {self.container.name!r}")
```

src/charm.py
benhoyt: If it can't happen, yeah, you're right. However, I believe charms can't set an error status, that's from Juju itself (if the hook can't be run, for example). So maybe if this "can't happen" you should check it further back, when `self._external_url` is being set, and just raise an exception so the charm goes into error state.

src/charm.py
```diff
@@ -428,7 +447,7 @@ def _common_exit_hook(self, update_ca_certs: bool = False) -> None:
         try:
             self.alertmanager_workload.update_config(self._render_manifest())
         except (ConfigUpdateFailure, ConfigError) as e:
-            self.unit.status = BlockedStatus(str(e))
+            self.last_error = str(e)
```
benhoyt, line 450: This is interesting. I can see why you're storing state in this attribute here (and below) in the common exit hook. However, normally you'd query in collect-status the current state of things, rather than look at previously-set state attributes. Might be tricky for this though?

benhoyt (commented):
Looks like a pretty good use of collect-status overall! Added a few comments to think about.

src/charm.py
sed-i: Agreed.

src/charm.py
sed-i: This is a leftover from when external url was a config option. Should be removed.

tests/unit/test_charm.py
sed-i: Yes!

sed-i (commented):
Thanks for reviewing @benhoyt and @dstathis!

I don't think this is ready to merge in light of the comments in the [lib PR](https://github.com/canonical/observability-libs/pull/63) since:
- we have custom events
- multiple entry points into a lib
- we shouldn't use collect_status outside of charm.py

My impression is that those constraints prevent us from a solid collect-status pattern.
I'm going to close this for now.

PR #211, opened 2024-01-27
Update libs, drop itest

sed-i:
Recently, we started seeing [failures](https://github.com/canonical/alertmanager-k8s-operator/actions/runs/7613950979/job/20899163062) in itests in CI.

## "config changed" error
This PR removes the `test_config_changed_modifies_file.py::test_update_config` test.

The test fetches the current config from the endpoint URL and the CI problem appeared because alertmanager started redacting the webhook url (now says `<secret>`):

```yaml
        webhook_configs:
        - http_config:
            enable_http2: true
            follow_redirects: true
          max_alerts: 0
          send_resolved: true
          url: <secret>
```

That corresponds with the error pebble surfaces:
```
caller=coordinator.go:118 level=error component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/alertmanager.yml err="scheme required for webhook url"
```

The test can be deleted because the same functionality is already captured in the `test_charm` unittest.

## TLS error
The CA cert was extracted using `juju show-unit ... ."relation-info"[0]`, with the `[0]` index being hard-coded.
Now that the ca charm has a "get ca cert" action, we get it directly from there.
I imagine the test was occasionally failing because the cert wasn't _always_ in relation "0".


PietroPasotti (approved):
fantastic! No idea how we'd all missed that <secret> in reldata.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1014,7 +1015,6 @@ def alerts(self) -> dict:
                 try:
                     scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
                     identifier = JujuTopology.from_dict(scrape_metadata).identifier
```
PietroPasotti, line 1017: we should move this down, out of the try/except block (it's not going to raise keyerror anyway)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1029,6 +1029,10 @@ def alerts(self) -> dict:
                 )
                 continue
 
+            # We need to append the relation info to the identifier. This is to allow for cases for there are two
+            # relations which eventually scrape the same application. Issue #551.
+            identifier = f"{identifier}_{relation.name}_{relation.id}"
```
PietroPasotti, line 1034: instead of redeclaring `'identifier'`, I'd create a new variable.

PietroPasotti (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: (This lib belongs to prom, not am.)

PietroPasotti: oh! silly me, didn't notice :)

PR #223, opened 2024-02-14
general maintenance and scenario tests fixes

PietroPasotti:
This PR:
- unpins ops since the bug in scenario that made that necessary is long gone
- bumps scenario to 6.0 
- fixes scenario tests
- bumps alertmanager_dispatch's dependency to pydantic v2
- fetches the latest ingress v2 lib which also uses pydantic v2
- pins pydantic>=2 in requirements.txt

## Testing Instructions
No charm code has been touched

### DIFF
--- a/lib/charms/alertmanager_k8s/v1/alertmanager_dispatch.py
+++ b/lib/charms/alertmanager_k8s/v1/alertmanager_dispatch.py
@@ -33,6 +33,7 @@ import pydantic
 from ops.charm import CharmBase, RelationEvent, RelationJoinedEvent, RelationRole
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Relation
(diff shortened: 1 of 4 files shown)


lib/charms/alertmanager_k8s/v1/alertmanager_dispatch.py
```diff
@@ -65,19 +66,21 @@ class _ProviderSchemaV1(pydantic.BaseModel):
     # TODO: when we switch to pydantic 2+, use computed_field instead of the following fields, and
     #  also drop the __init__.
     #  https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.computed_field
```
sed-i, line 68: Remove this?

sed-i (approved):

PR #249, opened 2024-05-03
make server cert path return None if tls disabled

PietroPasotti:
title says it all

charm tracing expects Optional[str], where None means: there's no TLS, and str means: here's your cert.
Alertmanager is returning str even though the file doesn't exist, which makes charm-tracing log exceptions on every hook.


sed-i (approved):

mmkay (approved):

PietroPasotti:
obsoleted by #251

PR #251, opened 2024-05-14
fix charm tracing cert

### LINKED ISSUE #250: charm tracing fails over tls
### Bug Description

When trying to integrate alertmanager with tempo to send its charm traces over tls, it fails to do with an exception "Exception while exporting Span batch."

### To Reproduce

juju deploy cos-lite --trust
juju refresh alertmanager --channel latest/edge
juju deploy tempo-k8s tempo --channel latest/edge
juju deploy self-signed-certificates
juju integrate self-signed-certificates:certificates alertmanager:certificates
juju integrate self-signed-certificates:certificates tempo:certificates
juju integrate tempo:tracing alertmanager:tracing
wait for an event on alertmanager (e.g: update-status)
look for error in juju logs

### Environment

juju version: 3.4.2-genericlinux-amd64
microk8s version: MicroK8s v1.28.7 revision 6532
alertmanager rev: 112

### Relevant log output

```shell
unit-alertmanager-0: 12:21:07 ERROR unit.alertmanager/0.juju-log Exception while exporting Span batch.
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/opentelemetry/sdk/trace/export/__init__.py", line 367, in _export_batch
    self.span_exporter.export(self.spans_list[:idx])  # type: ignore
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 145, in export
    resp = self._export(serialized_data)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 114, in _export
    return self._session.post(
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/requests/sessions.py", line 637, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/requests/adapters.py", line 458, in send
    self.cert_verify(conn, request.url, verify, cert)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/requests/adapters.py", line 261, in cert_verify
    raise OSError(
OSError: Could not find a suitable TLS CA certificate bundle, invalid path: /etc/alertmanager/alertmanager.cert.pem
```

### Additional context

_No response_

michaeldmitry:
## Issue
Fixes #250

## Solution
ca cert is pushed into charm container, so pass ca cert path to trace_charm decorator to use that for tls communication with tempo

## Context
1- Depends on https://github.com/canonical/tempo-k8s-operator/pull/98 
2- Assumes alertmanager and tempo have tls enabled
3- Assumes CA of alertmanager is the same for the one used by tempo's cert

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -56,7 +56,7 @@ logger = logging.getLogger(__name__)
 
 @trace_charm(
     tracing_endpoint="tracing_endpoint",
-    server_cert="server_cert_path",
+    server_cert="server_ca_cert_path",
     extra_types=(
         AlertmanagerProvider,
         CertHandler,
@@ -581,9 +581,9 @@ class AlertmanagerCharm(CharmBase):
         return None
 
     @property
-    def server_cert_path(self) -> Optional[str]:
-        """Server certificate path for tls tracing."""
-        return self._server_cert_path
+    def server_ca_cert_path(self) -> Optional[str]:
+        """Server CA certificate path for tls tracing."""
+        return self._ca_cert_path if self.server_cert.enabled else None
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -439,7 +439,7 @@ class AlertmanagerCharm(CharmBase):
             self.unit.status = BlockedStatus(str(e))
             return
 
-        if update_ca_certs:
+        if update_ca_certs and self.server_cert.enabled:
             self._update_ca_certs()
 
         # Update pebble layer
@@ -465,7 +465,7 @@ class AlertmanagerCharm(CharmBase):
 
     def _on_config_changed(self, _):
         """Event handler for ConfigChangedEvent."""
-        self._common_exit_hook()
+        self._common_exit_hook(update_ca_certs=True)
 
     def _on_peer_relation_joined(self, _):
         """Event handler for replica's RelationChangedEvent."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -439,7 +439,7 @@ class AlertmanagerCharm(CharmBase):
             self.unit.status = BlockedStatus(str(e))
             return
 
-        if update_ca_certs and self.server_cert.enabled:
+        if update_ca_certs:
             self._update_ca_certs()
 
         # Update pebble layer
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -383,7 +383,7 @@ def _render_manifest(self) -> ConfigFileSystemState:
                 self._templates_path: config_suite.templates,
                 self._amtool_config_path: config_suite.amtool,
                 self._server_cert_path: self.server_cert.server_cert,
-                self._key_path: self.server_cert.private_key,
+                self._key_path: self.server_cert.private_key if self.server_cert.enabled else None,
```
mmkay, line 386: Do we still need this change?

src/charm.py
```diff
@@ -439,7 +439,7 @@ def _common_exit_hook(self, update_ca_certs: bool = False) -> None:
             self.unit.status = BlockedStatus(str(e))
             return
 
-        if update_ca_certs:
+        if update_ca_certs and self.server_cert.enabled:
```
mmkay, line 442: This hook can also be triggered to unlink the certificates after the relation to CA is removed. With this change we wouldn't be unlinking the CA cert which we likely want to do.

mmkay (changes requested):
Not that up to date with alertmanager but I think we might be breaking the flow on unlinking the certificates here. 

Might be worth re-testing or adding a test case if that one isn't covered yet.

src/charm.py
michaeldmitry: yes
without it, private key will be generated and copied in the container filesytsem without even linking to certs
Also, UTs will break :D

mmkay: Should we do the same about the other paths then?

PietroPasotti: private_key is always present, it should just not be used as an existence check

PietroPasotti (approved):
LGTM.

mmkay (approved):

PR #264, opened 2024-06-10
add prometheus scrape jobs for all units

### LINKED ISSUE #151: Only leader scrape jobs are sent to Prometheus
### Enhancement Proposal

As @simskij said in [this comment](https://github.com/canonical/alertmanager-k8s-operator/pull/150#issuecomment-1513199310) Alertmanager is not providing a scrape job for all units, but only for the leader.

This is because Alertmanager requires `ingress-per-app`: 

```yaml
requires:
  ingress:
    interface: ingress
    limit: 1
    description: |
      Alertmanager typically needs a "per app" ingress, which is available in the traefik charm[1].

      [1] https://charmhub.io/traefik-k8s
```
We need `ingress-per-app` so alerts can be seen from outside the k8s cluster, but in the other hand we need `ingress-per-unit` so we can send unit scrapes jobs to Prometheus.

ca-scribner:
## Issue
closes #151

## Solution
Extends AlertmanagerCharm.self_scraping_job() to include all units, not just the leader, for metrics scraping.  

## Testing Instructions
```
juju deploy prometheus-k8s --channel edge --trust
charmcraft pack
juju deploy ./*.charm --resource alertmanager-image=ubuntu/alertmanager:0-22.04 --trust
juju relate alertmanager-k8s:self-metrics-endpoint prometheus-k8s

# wait for things to settle

jhack show-relation alertmanager-k8s prometheus-k8s
# Should see a single target provided

juju scale-application alertmanager-k8s 3
# wait for things to settle
jhack show-relation alertmanager-k8s prometheus-k8s
# Should see a target for each unit

# Check if prometheus has the targets
curl $PROMETHEUS_IP:9090/api/v1/targets \
  | jq '.data.activeTargets | .[] | {scrapeUrl, health}'
# should see a target for each am unit (plus one for prometheus)
```

## Upgrade Notes
Upgrade should happen seemlessly.  Nothing has changed to affect upgrades


src/alertmanager.py
```diff
@@ -106,6 +105,8 @@ def __init__(
         self._web_config_path = web_config_path
         self._is_tls_enabled = tls_enabled
 
+        self._peer_addresses = _parse_peer_addresses(peer_addresses, self._ha_port)
```
sed-i, line 108: `_get_peer_addresses` in `charm.py` is only used for this.
It renders addresses with the ha port, even though the ha port is passed as an arg to `WorkloadManager`.

Also, `web_external_url` is a deprecated config option so we do not expect to have a path component in the peer address.

So, I think we should refactor `_get_peer_addresses` to return the hostname. That could obviate the helpers added in this PR.

It would be great if we could use a consistent terminology for url parts, e.g. like [here](https://docs.python.org/3/library/urllib.parse.html).

ca-scribner: Looking at this a bit more, I'm not sure refactoring removes helpers so much as moves them around.

* `private_address` in the peer relation is `SCHEME://FQDN:API_PORT`
* scrape config tagets needs `FQDN:API_PORT` (no scheme)
* `WorkloadManager` needs `FQDN:HA_PORT` (no scheme)

So if we don't change `private_address`'s content, we still need helpers to remove scheme and change port.  

I've done another pass emphasizing reducing the code changes.  wdyt?  I think we still need some of these address conversions either way.

src/charm_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+        )
+    return parsed.hostname
+
+
+def add_port_to_addresses(addresses: List[str], port: Union[int, str]) -> List[str]:
+    """Adds a port to all the provided addresses.
+
+    Note that this does not check if the address already has a port.
+    """
+    return [f"{address}:{port}" for address in addresses]
```
sed-i, line 29: This is a trivial manipulation that seems to me to better have inlined.

src/charm.py
```diff
@@ -565,14 +566,17 @@ def _internal_url(self) -> str:
 
         If an external (public) url is set, add in its path.
```
sed-i, line 567:
```suggestion
```

src/charm.py
```diff
@@ -526,27 +529,25 @@ def _update_ca_certs(self):
             ca_cert_path.unlink(missing_ok=True)
         subprocess.run(["update-ca-certificates", "--fresh"], check=True)
 
-    def _get_peer_addresses(self) -> List[str]:
-        """Create a list of HA addresses of all peer units (all units excluding current).
+    def _get_peer_addresses(self, include_this_unit=True) -> List[str]:
+        """Returns a list of the addresses the peer units, without scheme or port.
```
sed-i, line 533: This is a good opportunity to start using correct terminology.

"address without scheme or port" is "hostname".

```mermaid
graph LR

scheme ---|://| username

subgraph netloc
username ---|:| password ---|@| hostname ---|:| port
end

port --- path["/path"]
```

image

tests/unit/test_scrape_jobs.py
sed-i, line 1: Should this be names "test_**self**_scrape_jobs"?

tests/unit/test_scrape_jobs.py
```diff
@@ -0,0 +1,81 @@
...
+        self.harness.set_leader(True)
+        self.app_name = "alertmanager-k8s"
+        # Create the peer relation before running harness.begin_with_initial_hooks(), because
+        # otherwise it will create it for you and we don't know the rel_id
+        self.peer_rel_id = self.harness.add_relation("replicas", self.app_name)
+
+        self.harness.begin_with_initial_hooks()
+
+    @patch.object(AlertmanagerCharm, "_internal_url", new_callable=PropertyMock)
+    @patch.object(AlertmanagerCharm, "_scheme", new_callable=PropertyMock)
```
sed-i, line 33: Can we have the values specified in the decorator, i.e. without using PropertyMock at all? [Example](https://github.com/canonical/alertmanager-k8s-operator/blob/eda1588bf24d6343b9ad32e96ad9506ed3028c90/tests/unit/test_charm.py#L22).

This change would also make the `jobs_expected` more explicit, which I think is a plus.
I.e. repetitions like this in tests are ok.

tests/unit/test_scrape_jobs.py
```diff
@@ -0,0 +1,81 @@
...
+    @patch("lightkube.core.client.GenericSyncClient")
+    @patch.object(WorkloadManager, "check_config", lambda *a, **kw: ("ok", ""))
+    @k8s_resource_multipatch
+    @patch.object(WorkloadManager, "_alertmanager_version", property(lambda *_: "0.0.0"))
+    def setUp(self, *unused):
+        self.harness = Harness(AlertmanagerCharm)
+        self.addCleanup(self.harness.cleanup)
+
+        self.harness.set_leader(True)
+        self.app_name = "alertmanager-k8s"
```
sed-i, line 25: Shorter is nicer to use in string literals.

```suggestion
        self.app_name = "am"
```

tests/unit/test_scrape_jobs.py
```diff
@@ -0,0 +1,81 @@
...
+            }
+        ]
+
+        # Add peers
+        for i, target in enumerate(targets[1:], 1):
+            unit_name = f"{self.app_name}/{i}"
+            self.harness.add_relation_unit(self.peer_rel_id, unit_name)
+            self.harness.update_relation_data(
+                self.peer_rel_id, unit_name, {"private_address": f"{scheme}://{target}"}
+            )
```
sed-i, line 78: We only add two units. I think it's better for clarity to be more explicit.

```suggestion
        self.harness.add_relation_unit(self.peer_rel_id, "am/1")
        self.harness.update_relation_data(self.peer_rel_id, "am/1", {"private_address": f"https://am-1.demo.internal:9093"})
        self.harness.add_relation_unit(self.peer_rel_id, "am/2")
        self.harness.update_relation_data(self.peer_rel_id, "am/2", {"private_address": f"https://am-2.demo.internal:9093"})
```
Same line count, but better clarity.

tests/unit/test_scrape_jobs.py
ca-scribner: afaik the linked method doesn't work for mocking properties, only methods.  I'll give it a shot though

tests/unit/test_scrape_jobs.py
ca-scribner: +1. done

tests/unit/test_scrape_jobs.py
ca-scribner: Something I'd like to avoid is having string literals that rely each other.  For instance, how `"am/1"` is written as a literal in a few places, and how the `targets` passed into the mock are written as separate literals from the `jobs_expected` literals.  Do you have a suggestion that feels better to you but also avoids this?

I took a second attempt at this, but I think it got worse... 
```python
    @patch.object(AlertmanagerCharm, "_internal_url", new_callable=PropertyMock)
    @patch.object(AlertmanagerCharm, "_scheme", new_callable=PropertyMock)
    def test_self_scraping_job_with_peers(self, _mock_scheme, _mock_internal_url):
        scheme = "https"
        _mock_scheme.return_value = scheme

        this_unit_internal_url = f"{scheme}://test-internal-0.url:1234"
        _mock_internal_url.return_value = this_unit_internal_url

        peer_unit_addresses = {
            "am/1": f"{scheme}://test-internal-1.url:1234",
            "am/2": f"{scheme}://test-internal-2.url:1234",
        }

        targets = [this_unit_internal_url] + [
            address.replace(f"{scheme}://", "") for address in peer_unit_addresses.values()
        ]

        jobs_expected = [
            {
                "metrics_path": "/metrics",
                "scheme": scheme,
                "static_configs": [{"targets": targets}],
            }
        ]

        # This could also be broken out like your first suggestion.  I think this is clearer, but happy to switch
        for unit_name, address in peer_unit_addresses.items():
            self.harness.add_relation_unit(self.peer_rel_id, unit_name)
            self.harness.update_relation_data(
                self.peer_rel_id, unit_name, {"private_address": address}
            )

        jobs_actual = self.harness.charm.self_scraping_job
        self.assertEqual(jobs_expected, jobs_actual)
```

src/charm.py
ca-scribner: Agreed.  [this commit](https://github.com/canonical/alertmanager-k8s-operator/pull/264/commits/8efdae9454016a51312dbe21babaf96ffa94aa54) I think covers the inconsistencies.  It reaches into the WorkloadManager as well

src/charm_helpers.py
ca-scribner: To me, the intent of this:

```python
        peer_ha_netlocs = add_port_to_hostnames(
            self._get_peer_hostnames(include_this_unit=False), self._ports.ha
        )
```

is more expressive than this:
```python
        peer_ha_netlocs = [
            f"{hostname}:{self._ports.ha}"
            for hostname in self._get_peer_hostnames(include_this_unit=False)
        ]
```

But, I dont feel strongly either way :D so I'll change it

src/charm_helpers.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Helper functions for the alertmanager charm."""
+from typing import List, Union
+from urllib.parse import urlparse
+
+
+def get_hostname_from_address(address: str) -> str:
```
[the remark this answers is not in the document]
ca-scribner: If `None`s are likely, I think we'd want to raise on them rather than filter them.  But looking at this again, I don't think they're likely (`None` happens if we `urlparse` on a netloc (eg: on something without a scheme)).  So what do you think of this?  

We can also filter on `None`, but I'm not sure we want to ignore `None` if it happens.  That feels like a sign of something badly wrong

sed-i (approved):

PR #273, opened 2024-07-16
add ingress refresh events to grafana source

### LINKED ISSUE #224: Set external URL as a Grafana Data Source
### Enhancement Proposal

Unlike Loki, if the relation between Alertmanager and Traefik is established, Alertmanager set fqdn address instead of the external one as a Grafana Data Source.

image

lucabello:
## Issue
Closes #224. Alertmanager is already using `external_url` in the Grafana Source integration. However, no refresh event is passed to `GrafanaSourceProvider`, meaning it won't start using the `external_url` if it doesn't have a traefik integration from the very start (e.g., when deployed from the bundle).

## Solution
Pass extra refresh events so the Grafana Source URL is updated correctly.

## Extra

`charmcraft fetch-lib`

## Additional Context
https://github.com/canonical/observability/blob/main/decision-records/2024-07-16--shared-urls.md

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -126,6 +126,7 @@ class AlertmanagerCharm(CharmBase):
             refresh_event=[
                 self.ingress.on.ready,
                 self.ingress.on.revoked,
+                self.on.update_status,
                 self.server_cert.on.cert_changed,
             ],
         )
--- a/src/charm.py
+++ b/src/charm.py
@@ -123,6 +123,11 @@ class AlertmanagerCharm(CharmBase):
             charm=self,
             source_type="alertmanager",
             source_url=self._external_url,
+            refresh_event=[
+                self.ingress.on.ready,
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -123,6 +123,11 @@ def __init__(self, *args):
             charm=self,
             source_type="alertmanager",
             source_url=self._external_url,
+            refresh_event=[
+                self.ingress.on.ready,
+                self.ingress.on.revoked,
+                self.server_cert.on.cert_changed,
+            ],
```
mmkay, line 130: In the [other PR](https://github.com/canonical/prometheus-k8s-operator/pull/632/files), you're also doing it on update-status: should we have it also in the list here?

lucabello: Good point :)

sed-i (approved):

PR #281, opened 2024-08-09
fix: add rust compiler for building charm dependencies

ca-scribner:
## Issue
fixes error in charmcraft file of "error: can't find Rust compiler", caused when building tracing library's pydanic dependency

## Solution
Add rustc and cargo to the charm build-packages

## Testing Instructions
If this charm packs successfully and passes integration tests, the fix worked

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -11,4 +11,8 @@ bases:
       channel: "20.04"
 parts:
   charm:
+    build-packages:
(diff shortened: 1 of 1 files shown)


sed-i (approved):

simskij:
While this is **technically** the right solution, the binary packages of pydantic and maturin resolves the need of having these packages in the first place. Closing the PR.

PR #286, opened 2024-08-28
test: fix deepdiff transient dep

dimaqq:
Same as https://github.com/canonical/prometheus-k8s-operator/pull/638


dimaqq:
An alternative would be to ask for `deepdiff<8`.

PietroPasotti (approved):

PR #329, opened 2025-04-17
tests: migrate to scenario 7

james-garner-canonical:
## Issue

Currently, `altertmanager-k8s-operator` specifies `scenario<7` for testing, due to breaking changes in the scenario 7 release. Charm Tech noticed this because we run this charm's unit tests in our CI against the latest versions of `ops` and `ops-scenario` in our CI to help us avoid breaking changes.

## Solution

This PR makes the minimal changes necessary to the unit tests to work with scenario 7, moving the tests in the separate `scenario` directory to the `unit` directory. It changes the dependency from `scenario<7` to `ops[testing]`. `ops[testing]` shouldn't use a qualifier since it is now released in lockstep with `ops`. After making this change, the dependencies have been re-locked via `tox -e lock`.

## Context

I'm not sure why making these changes requires us to mock more calls with `Exec` than there were originally. See comments below on specific lines -- hopefully these aren't problematic.

## Testing Instructions

Unit tests pass -- the same number of tests pass / skip / xfail now with `tox -e unit` as did on `main` with `tox -e unit -- tests/scenario`. There are now significantly more warnings though. There are some small changes in coverage, with total coverage going up slightly (~1%).

I'm also not sure why the `Check libraries` CI check is failing -- I assumed it was just failing on my fork.

## Upgrade Notes

There are a few small changes we'd recommend going forward, but which we left out of this PR to keep the diff down as they're more stylistic:
- pass sets to containers, relations, etc, rather than lists
- you should be able to avoid giving relations explicit IDs now
- consider `from ops import testing`, and then `testing.Relation` etc

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,4 +1,5 @@
 version = 1
+revision = 1
 requires-python = ">=3.8, <4"
 resolution-markers = [
     "python_full_version >= '3.11'",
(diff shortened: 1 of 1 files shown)


tests/unit/helpers.py
```diff
@@ -41,45 +42,49 @@ def begin_with_initial_hooks_isolated(context: Context, *, leader: bool = True)
...
-            ("update-ca-certificates", "--fresh"): ExecOutput(  # this is the command we're mocking
-                return_code=0,  # this data structure contains all we need to mock the call.
-                stdout="OK",
-            )
+        execs={
+            Exec(["update-ca-certificates", "--fresh"]),
+            Exec(
+                ["alertmanager", "--version"],
+                stdout="alertmanager, version 0.23.0 (branch: HEAD, ...",
+            ),
```
james-garner-canonical, line 50: The need for this additional mock is explained by a function being covered now when it wasn't before (`alertmanager.py` -> `WorkloadManager._alertmanager_version`) -- but I don't know why it's being called now when it wasn't previously.

tests/unit/helpers.py
```diff
@@ -41,45 +42,49 @@ def begin_with_initial_hooks_isolated(context: Context, *, leader: bool = True)
...
+            Exec(["/usr/bin/amtool", "check-config", "/etc/alertmanager/alertmanager.yml"]),
```
james-garner-canonical, line 51: I don't know why we need this one, as line 164-166 of `alertmanager.py` where this `Container.exec` takes place was apparently already covered on `main`.

lucabello:
Note: Check libraries fails because the automatically-generated `GITHUB_TOKEN` doesn't seem to have the correct permissions when it's from a fork.

PR #339, opened 2025-06-05
feat: service mesh support

adhityaravi:
## Feature

- Added support to add the alertmanager charm to the istio service service mesh

## Context

- The mesh relations for alertmanager were ingress and metrics endpoint. The ingress authz policy is automatically handled by relating to the [istio-ingress-k8s](https://github.com/canonical/istio-ingress-k8s-operator/) charm. The onyl authz policy that was added to the charm code is for exposing the metrics endpoint.

## Testing Instructions

- Pack the charm
```sh
charmcraft pack
```

- Deploy the stack
```sh 
# deploy istio core
juju add-model istio-system
juju deploy istio-k8s --trust --channel=latest/edge

# deploy alertmanager and friends
juju add-model dev
juju deploy istio-beacon-k8s --trust --channel=latest/edge
juju deploy istio-ingress-k8s --trust --channel=latest/edge
juju deploy prometheus-k8s --trust
juju deploy ./alertmanager-k8s_ubuntu@24.04-amd64.charm --trust --resource alertmanager-image=ubuntu/alertmanager:0-22.04

# wait for idle

# enable model on mesh
juju config istio-beacon-k8s model-on-mesh=true

# add alertmanager relations
juju integrate alertmanager-k8s:ingress istio-ingress-k8s:ingress-unauthenticated
juju integrate alertmanager-k8s istio-beacon-k8s
juju integrate alertmanager-k8s:self-metrics-endpoint prometheus-k8s:metrics-endpoint

# expose prometheus-k8s ui
kubectl port-forward --address=0.0.0.0 svc/prometheus-k8s 9090:9090 -n dev 
```

- Access the Prometheus UI and check the status of the alertmanager metrics provider

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -79,11 +79,18 @@ provides:
     optional: true
     description: |
       Configures Grafana to be able to use this Alertmanager instance as a datasource.
-  provide-cmr-mesh:
+  provide-cmr-mesh:  # server-side-for-cmr-mesh
     interface: cross_model_mesh
     description: |
-      Access a cross-model application from alertmanager via the service mesh.
-      This relation provides additional data required by the service mesh to enforce cross-model authorization policies
+      Allow cross-model applications to make HTTP requests to alertmanager via the service mesh.
+      This relation provides additional data required by the service mesh to create cross-model authorization policies
+
+      Announce a subset of juju topology to the other side because a CMR obfuscates identity.
+      Each pair of charm would need a separate relation of this kind, e.g. otelcol to loki and to prom.
+
+      To make use of this relation, you also must have either the service-mesh relation in place (e.g. istio-beacon) or
+      have the istio-beacon enroll the entire model (via its config option).
+      (The service_mesh charm library manages both of these relations.)
 
 requires:
   ingress:
@@ -118,8 +125,13 @@ requires:
     limit: 1
     interface: service_mesh
     description: |
-      Subscribe this charm into a service mesh to enforce authorization policies.
+      Subscribe this charm into a service mesh and create authorization policies.
+      We forward to the beacon our authorization policies.
+      The beacon sends the pod and service labels required by this charm to join the mesh.
+      This relation is a pre-requisite for using the provide-cmr-mesh relation.
   require-cmr-mesh:
+    # TODO: remove this relation when this is fixed:
+    #   https://github.com/canonical/istio-beacon-k8s-operator/issues/91
     interface: cross_model_mesh
     description: |
       Allow a cross-model application access to alertmanager via the service mesh.


charmcraft.yaml
```diff
@@ -79,6 +79,10 @@ provides:
     optional: true
     description: |
       Configures Grafana to be able to use this Alertmanager instance as a datasource.
+  provide-cmr-mesh:
+    interface: cross_model_mesh
+    description: |
+      Allows cross model access control via service mesh
```
sed-i, line 85: Can we make the description more insightful?
I have no clue what this means nor how it interops with the rest of the features of this charm.

Similar comment for the other descriptions.

src/charm.py
```diff
@@ -167,6 +168,22 @@ def __init__(self, *args):
...
 
+        self._mesh = ServiceMeshConsumer(
+            self,
+            policies=[
+                Policy(
+                    relation="self-metrics-endpoint",
+                    endpoints=[
+                        Endpoint(
+                            ports=[self.api_port],
+                            methods=[Method.get],
```
sed-i, line 179: (Separate issue) Should we have here `methods: Set[Literal["GET", ...]]` instead?

src/charm.py
```diff
@@ -167,6 +168,22 @@ def __init__(self, *args):
...
+                            paths=["/metrics"],
```
sed-i, line 180: The charms now rely on traefik doing stripPrefix.
Locally, I'd curl `localhost:9093/metrics`, but with ingress it would be `<trfk>/model-alertmanager/metrics` - different path and no port.
Are the two compatible?

charmcraft.yaml
adhityaravi: done!

src/charm.py
adhityaravi: Wasnt my decision, but I do prefer runtime validation over just static type checking. I see your point though, its only a couple of fixed nr. of options, Ill discuss with the team and get back.

src/charm.py
IbraAoad: So traefik isn't really compatible as an ingress provider when used in conjunction with istio, this policy will make sure that internal traffic, in this example `localhost:9093/metrics` is allowed from any charm that is related to `alertmanager` over the `self-metrics-endpoint` relation, in case of ingress we use the [istio-ingress](https://github.com/canonical/istio-ingress-k8s-operator) charm, which automatically generates appropriate Auth policies based on the routes requested by the ingressed charm.

@dstathis @ca-scribner feel free to jump in if my statement isn't correct or if you'd like to add more context

ca-scribner: You mean the strip prefix support which the charm requests through the ingress interface [here](https://github.com/canonical/alertmanager-k8s-operator/blob/c36c4491c48a2e14698008fa2ff14b242c7f9686/src/charm.py#L110)?  Yes, that'll be provided by istio just like traefik would provide it

ca-scribner: Re @IbraAoad's comment I think its slightly different to what @sed-i is saying.  I believe @sed-i is saying that when an outside user curls through the ingress,they'd curl `INGRESS/model-alertmanager/metrics` and it could get routed to:
1. (`strip_prefix=False`): `ALERTMANAGER:PORT/model-alertmanager/metrics`
2. (`strip_prefix=True`): `ALERTMANAGER:PORT/metrics`

In this case, @sed-i is saying his app wants to see the traffic like shown above in (2) (with `/model-alertmanager` removed).  I think this works the same way for both Traefik and an istio ingress, and in both cases we support the same options

src/charm.py
```diff
@@ -167,6 +168,22 @@ def __init__(self, *args):
 
         self.catalog = CatalogueConsumer(charm=self, item=self._catalogue_item)
 
+        self._mesh = ServiceMeshConsumer(
+            self,
+            policies=[
```
ca-scribner, line 173: afaik the [grafana-source](https://github.com/canonical/alertmanager-k8s-operator/blob/ccc9f5c31d210318de43534a04718928dc7d9951/charmcraft.yaml#L77-L81) relation will allow grafana's workload to GET (and maybe POST?) alertmanager's workload, so we'd need a policy for that

charmcraft.yaml
```diff
@@ -79,6 +79,11 @@ provides:
     optional: true
     description: |
       Configures Grafana to be able to use this Alertmanager instance as a datasource.
+  provide-cmr-mesh:
+    interface: cross_model_mesh
+    description: |
+      Access a cross-model application from catalogue via the service mesh.
```
sed-i, line 85: Copy-paste error?
```suggestion
      Access a cross-model application from alertmanager via the service mesh.
```

charmcraft.yaml
```diff
@@ -109,6 +114,16 @@ requires:
...
       Enables sending workload traces to a distributed tracing backend such as Tempo.
+  service-mesh:
+    limit: 1
+    interface: service_mesh
+    description: |
+      Subscribe this charm into a service mesh to enforce authorization policies.
+  require-cmr-mesh:
+    interface: cross_model_mesh
+    description: |
+      Allow a cross-model application access to catalogue via the service mesh.
```
sed-i, line 125: Copy-paste error?
```suggestion
      Allow a cross-model application access to alertmanager via the service mesh.
```

charmcraft.yaml
adhityaravi: yup, fixed.

charmcraft.yaml
adhityaravi: yup, fixed.

sed-i (approved):

PR #357, opened 2025-08-26
ci: provide a remote unit when departing

tonyandrewmeyer:
## Issue

When Juju emits a relation-departed event, there is always a `REMOTE_UNIT` set. However, this is not currently the case for one of the tests.

## Solution

Set the remote unit to an arbitrary unit ID (2) in the test.

## Context

In a future version of ops[testing], there will be a new consistency check for this case, to ensure that the event context is closer to the real context that Juju would provide. The test currently passes because ops[testing] (Scenario) lets it slide that one isn't there, but won't in that future version.

## Testing Instructions

Run the tests (or even just the specific test that's changing).

## Upgrade Notes

N/A

### DIFF
--- a/tests/unit/test_brute_isolated.py
+++ b/tests/unit/test_brute_isolated.py
@@ -15,7 +15,7 @@ def test_startup_shutdown_sequence(context: Context):
     state = context.run(context.on.update_status(), state)
 
     for peer_rel in state.get_relations("replicas"):
-        state = context.run(context.on.relation_departed(peer_rel), state)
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Hi @canonical/observability ! Is there something further I can do here? We'd like to get this fixed so that we can merge the change in Ops without any downstream failures.

Abuelodelanada (approved):
Looks reasonable. Thanks @tonyandrewmeyer !

Issue #316, opened 2025-02-07
Update datasource url on ingress changed/removed


michaeldmitry:
### Enhancement Proposal

Currently, in an event that the ingress url has changed (i.e Becomes a TLS endpoint), alertmanager doesn't listen to that event and update its grafana source provider URL.

Ref: https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py#L594

lucabello:
We should add the ingress events in the `refresh_events` of the instantiation of the `GrafanaSourceProvider`. Make sure that the `.ready` and `.revoked` custom events are also used.

lucabello:
Absolutely @Soundarya03 :) Feel free to submit a PR and ping me when you have something up, I'll jump on the review!

PR #365, opened 2025-09-17
fix: update grafana source url

Soundarya03:
fixes issue 316 (Update datasource url on ingress changed/removed)

## Issue
This is a fix for #316 

## Solution
The `update_source` method is invoked with the external url on the common exit hook, which is triggered on ingress ready and revoked events, besides others. The [prometheus-k8s charm's](https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py) implementation of the same has been taken as a reference.

## Testing Instructions

These were the steps taken to test the change: 

1. Pack the charm: `charmcraft pack`

2. Deploy it: `juju deploy ./alertmanager-k8s.charm --resource alertmanager-image=ubuntu/alertmanager:0-22.04`

3. Deploy traefik: `juju deploy traefik-k8s`

4. Integrate over ingress: `juju relate alertmanager-k8s:ingress traefik-k8s:ingress`

5. Verify external url provisioned: `juju show-unit alertmanager-k8s/0'

6. Modify this url: `juju config traefik-k8s external_hostname=alertmanager.local`

7. Verify url once more as in step 5

8. Deploy grafana: `juju deploy grafana-k8s --channel 1/stable --trust`

9. Integrate grafana with alertmanager: `juju relate alertmanager-k8s:grafana-source grafana-k8s:grafana-source`

10. Verify grafana source url: `juju show-unit grafana-k8s/0`

11. Modify url as in step 6.

12. Verify grafana source url once more as in step 9. Without this fix, it should still be holding the old url. With the fix, we should see that it is updated as expected.


lucabello:
Hey @Soundarya03, that does look like something we missed, so I opened https://github.com/canonical/prometheus-k8s-operator/issues/728 for us to investigate further :) Nice job spotting that!

PR #392, opened 2026-02-11
chore: Bumps Alertmanager to 0.31

Gmerold:
Bumps Alertmanager to `0.31`


lucabello (approved):
This PR is simply speeding up the Renovate pr (which would run weekly) to pull the new version of Alertmanager into the charm, which we need for our PS7 deployment.

PR #403, opened 2026-04-21
feat(terraform): add channel validation

lucabello:


MichaelThamm:
See this PR which removed validation in the products:
- https://github.com/canonical/observability-stack/pull/274/changes

You can copy out whatever you think makes sense there e.g., the unit tests to assert that the correct track is being supported. However, this seems to provide little value. The margin for error will be highest when we cut new tracks end of cycle and update this field from `dev` to `x.y`, maybe just hardcoding without utest is fine.

MichaelThamm (approved):

PR #408, opened 2026-04-29
feat: migrate charm-tracing to ops[tracing]

lucabello:
## Summary

Migrate charm tracing from the old `charm_tracing` charm library (`@trace_charm` decorator) to the new `ops[tracing]` approach using `ops_tracing.Tracing`.

### Changes
- Replaced `@trace_charm` decorator with `ops_tracing.Tracing` instantiation
- Updated dependencies to use `ops[tracing]`
- Added `cargo` to build-packages (pydantic-core dependency)
- Removed old `charm_tracing.py` library (kept `tracing.py` for workload-tracing)
- Updated test fixtures for new tracing approach

### Motivation
The `ops[tracing]` approach is the recommended way to add charm tracing as of ops 2.20.0.


src/charm.py
```diff
@@ -184,8 +173,10 @@ def __init__(self, *args):
             ],
         )
         self._tracing = TracingEndpointRequirer(self, protocols=["otlp_http"])
-        self._charm_tracing_endpoint, self._charm_tracing_ca_cert = charm_tracing_config(
-            self._tracing, self._ca_cert_path
+        self.tracing = ops_tracing.Tracing(
```
mmkay: Code readability thing: now we have `self._tracing` and `self.tracing`. One is also private, the other isn't. Would it make sense to call them `workload_tracing` and `charm_tracing` and choose one convention for all?

mmkay (approved):
One important note, but other than that we're fine :)

src/charm.py
lucabello: Good point!

charmcraft.yaml
```diff
@@ -135,6 +135,18 @@ requires:
     limit: 1
     description: |
       Enables sending workload traces to a distributed tracing backend such as Tempo.
+  charm-tracing:
+    interface: tracing
+    optional: true
+    limit: 1
+    description: |
+      Enables sending charm traces to a distributed tracing backend such as Tempo.
+  receive-ca-cert:
```
sed-i, line 144: Fetch-lib, etc. for recv-ca-cert?

Abuelodelanada (approved):
LGTM

charmcraft.yaml
lucabello: Done!

PR #434, opened 2026-06-19
feat: non-root user compliance

deusebio:
## Assessment
| Component | Status | Details |
|-----------|--------|---------|
| **Charm container** | ❌ Non-compliant | `charm-user: non-root` was not set in `charmcraft.yaml` |
| **Workload container (alertmanager)** | ❌ Non-compliant | `uid` and `gid` were not set to `584792` |
| **Upstream image permissions** | ❌ Non-compliant | Upstream image (`ubuntu/alertmanager`) runs as `root` (uid=0); all config paths owned by root |
| **Pebble operations** | ✅ Paths accessible | Charm writes to `/etc/alertmanager/`, `/etc/amtool/`, `/usr/local/share/ca-certificates/`, `/etc/ssl/certs/` |

## Mitigation
- Added `charm-user: non-root` to `charmcraft.yaml`
- Set `uid: 584792` and `gid: 584792` for the `alertmanager` container in `charmcraft.yaml`
- Rebuilt the alertmanager image (`0.31.1`) from source using `rockcraft` with:
  - `run_user: _daemon_` to set the runtime user to `_daemon_` (uid=584792, gid=584792)
  - A `user-setup` part to set correct ownership and permissions on `/etc/alertmanager/`, `/etc/amtool/`, `/usr/local/share/ca-certificates/`, `/etc/ssl/certs/`
- Published new image to `edeusebio85/alertmanager:0.31.1-non-root`
- All 6 non-root image tests pass locally

## Verification
- Added `test_container_security_context` parametrized integration test in `tests/integration/test_kubectl_delete.py`
- Added helper functions (`generate_container_securitycontext_map`, `get_pod_names`, `assert_security_context`) to `tests/integration/helpers.py`

## Related

- Rock image PR (rockcraft changes): canonical/alertmanager-rock#64


charmcraft.yaml
```diff
@@ -180,7 +184,7 @@ resources:
       OCI image for alertmanager. This charms makes the following assumptions about the image:
       - location of executable "alertmanager" is in the path
       - has `update-ca-certificates`
-    upstream-source: ubuntu/alertmanager@sha256:6d137784294d137394683ec56519eafce6d0bd91c4c62439197bb1c898d1a76e  # renovate: oci-image tag: 0.31-26.04
+    upstream-source: edeusebio85/alertmanager:0.31.1-non-root
```
lucabello, line 187: Reminder to revert this after we merge the rock

charmcraft.yaml
```diff
@@ -33,7 +33,7 @@ assumes:
   - juju >= 3.6
 
 platforms:
-  ubuntu@26.04:amd64:
+  ubuntu@24.04:amd64:
```
lucabello, line 36: Why was this reverted to 24.04? We should turn it back to 26.04

pyproject.toml
```diff
@@ -3,7 +3,7 @@
 [project]
 name = "alertmanager-k8s"
 version = "0.0"
-requires-python = "~=3.14.0"
+requires-python = "~=3.12.0"
```
lucabello, line 6: Reverting the base to 26.04 will mean reverting the Python version to 3.14 instead of 3.12 as well

charmcraft.yaml
deusebio: Yes, good point. There is an issue with pebble and juju in 26.04. I have flagged this by Juju team and ops, see [here](https://matrix.to/#/!EfnrLPl1aPaEUQlLDDgKHtOfvsx6xGDUkel5gim8l0U/$GMYLmkWNKdLjRmYCxqwWogXEhyJytRmQzHWrr4jj1y0?via=ubuntu.com&via=matrix.org&via=laquadrature.net). 

I can also revert it back, but I would *personally* do this once Charm Tech team releases the fix

charmcraft.yaml
deusebio: Yes, this depends on merging https://github.com/canonical/alertmanager-rock/pull/64. Let's leave this comment open until the other PR is merged and this can be upgraded to the correct reference.

pyproject.toml
deusebio: Yes, indeed! I'll resolve this, as it is already included as part of [this](https://github.com/canonical/alertmanager-k8s-operator/pull/434#discussion_r3490075184)

src/charm.py
```diff
@@ -223,6 +225,8 @@ def __init__(self, *args):
             for hostname in self._get_peer_hostnames(include_this_unit=False)
         ]
 
+        ca_cert_path = Path(self._local_cert_path) / Path(self._ca_cert_path).name
```
sed-i, line 228: This confused me a bit, should we name it differently?

```suggestion
        local_ca_cert_path = Path(self._local_cert_path) / Path(self._ca_cert_path).name
```

src/charm.py
```diff
@@ -648,7 +652,7 @@ def _update_ca_certs(self):
         # Workload container
         self.container.exec(["update-ca-certificates", "--fresh"], timeout=30).wait()
         # Charm container
-        subprocess.run(["update-ca-certificates", "--fresh"], check=True)
+        subprocess.run(["sudo","update-ca-certificates", "--fresh"], check=True)
```
sed-i, line 655: Iiuc, this `update-ca-cetificates` will not pick up the certs we are now placing in `_local_cert_path = "./certs"`.
We want those certs (or at least our own cert) in the cert store because the charm makes client calls to the workload, that could be behind TLS.
Not sure if this is still the case.

@sinapah I see you've been contributing here recently, perhaps you have better context?

tests/integration/test_kubectl_delete.py
```diff
@@ -33,6 +44,20 @@ def test_deploy(juju, charm_path: Path):
     )
 
 
+@pytest.mark.parametrize("container_name", list(CONTAINERS_SECURITY_CONTEXT_MAP.keys()))
+def test_container_security_context(juju, container_name: str):
```
sed-i, line 48: Can we simplify this a bit? We don't need to test that charmcraft is doing what it's supposed to do.
I think this can be reduced to a one-liner yq assertion such as

```bash
yq -e ".containers | map(.uid == 584792 and .gid == 584792) | all" charmcraft.yaml
```

charmcraft.yaml
sed-i: Let's wait for the fix, we can't flip this back to 24.04 all of a sudden.

sed-i (commented):
Thanks @deusebio !
Some comments below.

src/charm.py
sinapah: We do indeed make client calls from the charm to the workload. For example,
https://github.com/canonical/alertmanager-k8s-operator/blob/e572d44cd36fb431b11a37238932254b35fb6634/src/alertmanager_client.py#L86-L114
and 
https://github.com/canonical/alertmanager-k8s-operator/blob/e572d44cd36fb431b11a37238932254b35fb6634/src/alertmanager_client.py#L42-L55

sinapah: This is potentially a no-op, but the client calls from the charm to the workload should still be safe?
The PR does correctly do
```py
cafile=str(ca_cert_path) if ca_cert_path.exists() else None,
```
so I believe the reload and status calls should still succeed. I think the fact that the `test_tls_web` itest is succeeding is a positive sign about this.

deusebio: Yes, the charm is writing the CA certs (to be used in HTTPS calls against the workload) in the `./certs` folder that is accessible by the charm, and those certs are then passed to the "WorkloadManager" class that is making the calls from the charm.

So the current implementation seems sensible and sane, but if the `"update-ca-certificates", "--fresh"` is not needed, we can probably remove this line and also allow the charm to run as `non-root`. The reason why we set `charm-user: sudoer` was exactly because of this line

deusebio: I have raised a PR to give this a go https://github.com/canonical/alertmanager-k8s-operator/pull/451 

~~https://github.com/deusebio/alertmanager-k8s-operator/pull/1/changes~~

*edit: the ci was not clean on the fork*

charmcraft.yaml
deusebio: This is fixed in ops 3.8.0, see [here](https://github.com/canonical/operator/releases/tag/3.8.0) (that includes https://github.com/canonical/operator/issues/2547)

Should we bump ops from 2.x to 3.x in the context of this PR?

PR #437, opened 2026-06-26
feat: loosely AlertmanagerPagerdutyNotificationsFailing 

gabrielcocenza:
The current alert expression:

```promql
rate(alertmanager_notifications_failed_total{integration=".*"}[5m]) > 0
```

Is the noisiest alert in the stack. Any single failed notification attempt — including transient TCP connection rejections — triggers it. This causes near-constant firing that desensitizes on-call engineers and reduces trust in the alert.

The proposed change is:
```promql
expr: increase(alertmanager_notifications_failed_total[5m]) > 5
```

Meaning: If more than 5 Alertmanager notification delivery failures (across any integration) occur within a trailing 5-minute window, and that condition holds continuously for 10 minutes, fire a warning alert.

### DIFF
--- a/src/prometheus_alert_rules/alertmanager_notifications_failed.rule
+++ b/src/prometheus_alert_rules/alertmanager_notifications_failed.rule
@@ -2,7 +2,7 @@ groups:
 - name: AlertmanagerNotificationsFailed
   rules:
   - alert: AlertmanagerNotificationsFailed
-    expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0
+    expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0.02
     for: 0m
     labels:
       severity: warning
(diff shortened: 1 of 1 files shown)


src/prometheus_alert_rules/alertmanager_notifications_failed.rule
```diff
@@ -2,7 +2,7 @@ groups:
 - name: AlertmanagerNotificationsFailed
   rules:
   - alert: AlertmanagerNotificationsFailed
-    expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0
+    expr: rate(alertmanager_notifications_failed_total{integration=~".*"}[5m]) > 0.02
```
sed-i: For future reference, could you please add as a comment here what 0.02 stands for?
I imagine:

$$
\left.\frac{\Delta N_f}{\Delta t}\right|_{5\mathrm{min}} = \frac{6}{5\cdot 60} = 0.02 \cdot \mathrm{s}^{-1}
$$

I.e. 6 failures in 5 minutes on average.

Note that it could be easier to using `increment` in this case instead of `rate`.

gabrielcocenza: Agree that increment is more human palatable. Changed to:

```promql
expr: increase(alertmanager_notifications_failed_total[5m]) > 5
```

Also increased the for to 10m

sed-i (approved):
Thank you!

